PR #1, opened 2022-12-06
feat: app boilerplate

mhuynh5757:
**Contributing to Twilio**

This PR replaces the former example app with a brand-new templated boilerplate using the React Native CLI.


charliesantos:
@mhuynh5757 there's already some files in the [task/monorepo-restructure](https://github.com/twilio/twilio-voice-react-native-app/tree/task/monorepo-restructure) branch, the one you're submitting this PR to. I think we need to review those changes first before submitting more PRs against it. Can you please submit a PR to main branch instead so we know the source of truth?

app/.node-version
```diff
@@ -0,0 +1 @@
+16
```
charliesantos, line 1: Can we use the latest LTS version? I think we're at 18 now.

app/.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  arrowParens: 'avoid',
+  bracketSameLine: true,
+  bracketSpacing: false,
+  singleQuote: true,
+  trailingComma: 'all',
```
charliesantos, line 6: Is this the default, or is this what you're proposing? I prefer `bracketSpacing` to be true...

app/.eslintrc.js
```diff
@@ -0,0 +1,16 @@
...
+  extends: '@react-native-community',
+  parser: '@typescript-eslint/parser',
+  plugins: ['@typescript-eslint'],
+  overrides: [
+    {
+      files: ['*.ts', '*.tsx'],
+      rules: {
+        '@typescript-eslint/no-shadow': ['error'],
+        'no-shadow': 'off',
+        'no-undef': 'off',
```
charliesantos, line 12: Is this just default or what you're proposing? I think we need to add more rules. Probably not in this PR but wanted to know your intention.

charliesantos (commented):
Left a few comments. We need some eyes on the android and ios related parts. @kbagchiGWC @bobiechen-twilio .

Also, please update all references of "TwilioVoiceReactNativeReferenceApp" to "TwilioVoiceReactNativeApp". That is, no need to mention the "reference" part.

app/.node-version
mhuynh5757: This is recommended by the React Native team, but I can try with 18 from here on out. If nothing goes wrong I'll change this in future PRs.

app/.prettierrc.js
mhuynh5757: I'll change this to `true`, it was `false` by default from the RN boilerplate.

app/.eslintrc.js
mhuynh5757: This file is entirely defaults, let's change things up as we see fit. 👍

app/.watchmanconfig
```diff
@@ -0,0 +1 @@
+{}
```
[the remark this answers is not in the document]
mhuynh5757: Default file included with the RN boilerplate, not sure what happens if we remove it.

PR #13, opened 2023-02-22
VBLOCKS-1438: refactor - remove rudimentary auth

kpchoy:
[VBLOCKS-1438](https://issues.corp.twilio.com/browse/VBLOCKS-1438)

Remove existing rudimentary auth

**Contributing to Twilio**


charliesantos (changes requested):
@kpchoy I see a lot of changes updating single quotes to double quotes. Is this from your local linter? Please change it back first then I'll continue to review.

kpchoy:
@charliesantos sorry about that, needed to add a prettierrc.js to server folder

server/.prettierrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  bracketSameLine: true,
+  singleQuote: true,
+  trailingComma: "all",
```
charliesantos, line 4:
```suggestion
  trailingComma: 'all',
```

mhuynh5757, line 4:
```suggestion
  trailingComma: 'all',
```
Super small nitpick, just to match the `prettier` config from the `app/` folder.

charliesantos (changes requested):
I still see references of auth related code in the following areas. Do we need to remove these or are we thinking of using them later? @kpchoy @mhuynh5757 
- app/src/store/user.ts and store.test.ts
- token.test.ts
- twiml.test.ts
- server.test.ts

app/__tests__/store.test.ts
```diff
@@ -38,12 +38,6 @@ it('works good', async () => {
   console.log(app.store.getState());
 
   fetchMock.mockResolvedValueOnce({ status: 200 });
```
mhuynh5757, line 40:
```suggestion
  
```
We shouldn't need this line or the following `console.log` anymore.

server/src/routes/token.ts
```diff
@@ -1,27 +1,19 @@
...
-      res.status(403).send('Unauthenticated request.');
-      return;
-    }
-
     const accessToken = new AccessToken(
       twilioCredentials.ACCOUNT_SID,
       twilioCredentials.API_KEY_SID,
       twilioCredentials.API_KEY_SECRET,
-      {
-        identity: authentication.username,
```
mhuynh5757, line 23: Note for future reference, we need this for incoming but are removing it presently just for finishing outgoing.

mhuynh5757 (approved):
Looking good!

mhuynh5757:
> I still see references of auth related code in the following areas. Do we need to remove these or are we thinking of using them later? @kpchoy @mhuynh5757
> 
> * app/src/store/user.ts and store.test.ts
> * token.test.ts
> * twiml.test.ts
> * server.test.ts

For the test stuff, I think we can revisit when we do test tooling.

server/src/routes/token.ts
kpchoy: Should I just comment it out and leave a note here @mhuynh5757 ?

mhuynh5757: Nah, I think it's ok to remove it. We don't know how the final API will end up looking, i.e. what if we have a different route for getting tokens for incoming calls.

charliesantos (approved):
Ok to merge once @mhuynh5757 's comments are addressed.

PR #15, opened 2023-03-01
Feature/protect token route

kpchoy:
https://issues.corp.twilio.com/browse/VBLOCKS-1460

Protect Token Route with auth0 authentication

**Contributing to Twilio**

### DIFF
--- a/app/config.ts
+++ b/app/config.ts
@@ -2,6 +2,7 @@ const config = {
   clientId: '{ClientID}',
   domain: '{DomainName}',
   auth0Scope: 'openid profile email',
+  audience: '{yourApiIdentifier}',
 };
 
 export default config;
--- a/app/config.ts
+++ b/app/config.ts
@@ -1,7 +1,22 @@
+/**
+ * Auth0 related configurations
+ */
 const config = {
+  /**
+   * The clientId and domain for your Auth0 (native) application.
+   * See https://auth0.com/docs/get-started/applications/application-settings#basic-settings
+   */
   clientId: '{ClientID}',
   domain: '{DomainName}',
+  /**
+   * The Open ID Connect Scopes for your Auth0 (native) application.
+   * See https://auth0.com/docs/get-started/apis/scopes/openid-connect-scopes
+   */
   auth0Scope: 'openid profile email',
+  /**
+   * The Identifier field on your Auth0 (server) API's settings tab.
+   * See https://auth0.com/docs/secure/tokens/access-tokens/get-access-tokens
+   */
   audience: '{yourApiIdentifier}',
 };
 
--- a/app/src/screens/SignIn/index.tsx
+++ b/app/src/screens/SignIn/index.tsx
@@ -2,6 +2,8 @@ import React from 'react';
 import { useAuth0 } from 'react-native-auth0';
 import { StyleSheet, View, Button, Image, Text } from 'react-native';
 import config from '../../../config';
+import { useDispatch } from 'react-redux';
+import { setAccessToken } from '../../store/user';
 
 const TwilioLogo = require('../../../assets/icons/logo-twilio-red.png');
 const HelloFigure = require('../../../assets/icons/hello-figure.png');
@@ -38,15 +40,22 @@ const styles = StyleSheet.create({
 });
 
 const SignIn: React.FC = () => {
-  const { authorize } = useAuth0();
+  const { authorize, getCredentials } = useAuth0();
+  const dispatch = useDispatch();
 
   const onLogin = async () => {
     try {
-      await authorize({ scope: config.auth0Scope });
+      await authorize({
+        scope: config.auth0Scope,
+        audience: config.audience,
+      });
+      const { accessToken } = await getCredentials();
+      dispatch(setAccessToken(accessToken));
     } catch (e) {
       console.error(e);
     }
   };
+
   return (
     <View style={styles.container}>
       <View style={styles.logoContainer}>
--- a/app/src/store/user.ts
+++ b/app/src/store/user.ts
@@ -1,0 +1,20 @@
+import { createSlice } from '@reduxjs/toolkit';
+import type { PayloadAction } from '@reduxjs/toolkit';
+
+interface UserState {
+  accessToken: string;
+}
+
+const initialState = { accessToken: '' } as UserState;
+
+export const userSlice = createSlice({
+  name: 'user',
+  initialState,
+  reducers: {
+    setAccessToken: (state, action: PayloadAction<string>) => {
+      state.accessToken = action.payload;
+    },
+  },
+});
+
+export const { setAccessToken } = userSlice.actions;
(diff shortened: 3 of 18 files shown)


mhuynh5757:
Do we have a unit test for the unhappy route in `server.test.ts`? Meaning when the middleware would send a `401`.

kpchoy:
> Do we have a unit test for the unhappy route in `server.test.ts`? Meaning when the middleware would send a `401`.

Not yet, will work on adding

charliesantos:
@kpchoy this PR is showing changes that are already in the public beta branch. Can you please create a separate PR for updating [feature/3rd_party_auth](https://github.com/twilio/twilio-voice-react-native-app/tree/feature/3rd_party_auth) branch, review and merge that first, then we can review this current PR. That way, we can only see what you changed.

kpchoy:
> @kpchoy this PR is showing changes that are already in the public beta branch. Can you please create a separate PR for updating [feature/3rd_party_auth](https://github.com/twilio/twilio-voice-react-native-app/tree/feature/3rd_party_auth) branch, review and merge that first, then we can review this current PR. That way, we can only see what you changed.

Sure will do that

app/src/screens/SignIn/index.tsx
```diff
@@ -38,15 +40,22 @@ const styles = StyleSheet.create({
...
+  const dispatch = useDispatch();
 
   const onLogin = async () => {
     try {
-      await authorize({ scope: config.auth0Scope });
+      await authorize({
+        scope: config.auth0Scope,
+        audience: config.audience,
+      });
+      const { accessToken } = await getCredentials();
```
mhuynh5757, line 52: I think this sort of cascading action logic should belong in the store/actions file. Let's think of an async action that performs this sort of authentication -> token fetching logic in one action we can dispatch from this component.

mhuynh5757: Actually, I can see the merit of two separate actions here. Let's move the authorize call into an action, though.

mhuynh5757: Does Auth0 have utility functions for this sort of functionality outside of React hooks? Could we fabricate our own actions instead?

app/src/store/user.ts
```diff
@@ -0,0 +1,20 @@
+import { createSlice } from '@reduxjs/toolkit';
+import type { PayloadAction } from '@reduxjs/toolkit';
+
+interface UserState {
+  accessToken: string;
+}
+
```
mhuynh5757, line 7: Ideally, this is where I'd like all the authentication functionality to behave, i.e. calls to Auth0, getting Auth0 tokens, etc.

server/src/server.ts
```diff
@@ -3,18 +3,25 @@ import type { TwilioCredentials } from './common/types';
 import { createTokenRoute } from './routes/token';
 import { createTwimlRoute } from './routes/twiml';
 import { createLogMiddleware } from './middlewares/log';
+import { auth } from 'express-oauth2-jwt-bearer';
+import config from '../config';
 
 export function createExpressApp(twilioCredentials: TwilioCredentials) {
```
mhuynh5757, line 9: We should think of some way to incorporate the auth configs into a similar mechanism. Maybe we rename from `twilioCredentials` to `serverConfig` or something similar, but I don't think we should bake configs into more than one file or mechanism (env vars).

app/config.ts
```diff
@@ -2,6 +2,7 @@ const config = {
   clientId: '{ClientID}',
   domain: '{DomainName}',
   auth0Scope: 'openid profile email',
+  audience: '{yourApiIdentifier}',
```
charliesantos, line 5: What is this new config for?
We should probably add docstrings for each of this field.

server/config.ts
```diff
@@ -0,0 +1,6 @@
+const config = {
+  audience: '{yourApiIdentifier}',
+  issuerBaseURL: `https://{yourDomain}/`,
```
charliesantos, line 3: Do we expect customers to update this file? If so, these values should go to environment variables.

app/src/store/voice/token.ts
```diff
@@ -6,11 +6,13 @@ export const getToken = createAsyncThunk<
...
   void,
   { state: State; dispatch: Dispatch }
->('voice/getToken', async () => {
+>('voice/getToken', async (_, thunkAPI) => {
+  const accessToken = thunkAPI.getState().voice.user.accessToken;
   const res = await fetch(`${defaultUrl}/token`, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
+      authorization: `Bearer ${accessToken}`,
```
charliesantos, line 15:
```suggestion
      Authorization: `Bearer ${accessToken}`,
```

server/src/server.ts
```diff
@@ -3,18 +3,25 @@ import type { TwilioCredentials } from './common/types';
...
+
   app.use(express.urlencoded({ extended: false }));
   app.use(express.json());
 
   app.use(createLogMiddleware());
 
   const tokenRouter = Router();
   tokenRouter.use(createTokenRoute(twilioCredentials));
-  app.post('/token', tokenRouter);
+  app.post('/token', jwtCheck, tokenRouter);
```
charliesantos, line 24: Oh this is cool! This means any endpoint that we want to protect, we just do this same thing?

charliesantos (changes requested):

app/config.ts
kpchoy: What kind of docstring do you have in mind? Reference link to auth0?

server/src/server.ts
kpchoy: Yup we would just need to add `jwtcheck` to each route we want to protect

app/src/screens/SignIn/index.tsx
mhuynh5757: That's a good point. My opinion is that this kind of logic seems very "atomic" in the sense we don't really need to use the steps individually anywhere in the app. The hook they provide seems more for apps that don't use centralized logic, but I can see your point.

app/config.ts
charliesantos: Something like the following. I added an example for clientId. Just fill in the docstrings for the rest.
```ts
/**
 * Auth0 related configurations
 */
const config = {
  /**
   * The unique identifier for your Auth0 application.
   * See https://auth0.com/docs/get-started/applications/application-settings#basic-settings
   */
  clientId: '{ClientID}',
  domain: '{DomainName}',
  auth0Scope: 'openid profile email',
  audience: '{yourApiIdentifier}',
}
```

charliesantos: I understand we don't do this in most part of the code. But I think we should start doing it on areas where things are not self explanatory. For example on this new config, I'm not sure what audience really means :D

mhuynh5757 (commented):
The server config stuff is looking good!

server/src/server.ts
charliesantos: Not for this PR, but is it possible to get other user information such as email or username from the jwt? Or can you post an example jwt?

charliesantos (approved):
Looks really good. +1 for me. Let's wait for @mhuynh5757 's +1 before merging.

server/src/server.ts
charliesantos: Ok, we can take a look at that later. But can you please post an example jwt here? I like to see it's structure.

PR #24, opened 2023-03-09
[VBLOCKS-1489] Refactor auth0 wrapper

kpchoy:
[VBLOCKS-1489](https://issues.corp.twilio.com/browse/VBLOCKS-1489)

Wrap Auth0 in Redux actions. 

**Contributing to Twilio**


app/src/screens/SignIn/index.tsx
```diff
@@ -40,20 +39,15 @@ const styles = StyleSheet.create({
...
-        audience: config.audience,
-      });
-      const { accessToken } = await getCredentials();
-      dispatch(setAccessToken(accessToken));
-    } catch (e) {
-      console.error(e);
-    }
+  React.useEffect(() => {
+    dispatch(checkLoginStatus());
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
mhuynh5757, line 46: Let's leave this out, I think it's fine to include dependencies in the effect hook that shouldn't change. I'm assuming it's `dispatch` that it's concerned might change and I get that it will never change, but I think it's fine to just include it.

app/src/screens/Home/index.tsx
```diff
@@ -1,12 +1,14 @@
...
-import { useAuth0 } from 'react-native-auth0';
 import {
   StyleSheet,
   Text,
   View,
   Image,
   TouchableHighlight,
 } from 'react-native';
+import { useDispatch, useSelector } from 'react-redux';
+import { Dispatch, type State } from '../../store/app';
```
mhuynh5757, line 10:
```suggestion
import { type Dispatch, type State } from '../../store/app';
```

app/src/store/voice/token.ts
```diff
@@ -6,23 +6,26 @@ export const getToken = createAsyncThunk<
...
-    }),
-  });
+>('voice/getToken', async (_, { getState }) => {
+  const user = getState().voice.user;
+  if (user?.status === 'fulfilled') {
+    const res = await fetch(`${defaultUrl}/token`, {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Authorization: `Bearer ${user?.accessToken}`,
```
mhuynh5757, line 16:
```suggestion
        Authorization: `Bearer ${user.accessToken}`,
```
Checking if `user?.status === 'fulfilled'` should type-guard this for us.

mhuynh5757 (changes requested):

app/src/store/user.ts
```diff
@@ -1,20 +1,107 @@
...
+  const credentials = await auth0.webAuth.authorize({
+    scope: config.auth0Scope,
+    audience: config.audience,
+  });
+
+  if (typeof credentials.idToken === 'undefined') {
+    return rejectWithValue('ID_TOKEN_UNDEFINED');
+  }
+
+  auth0.credentialsManager.saveCredentials(credentials as any);
```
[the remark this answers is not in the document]
kpchoy: Screen Shot 2023-03-13 at 4 17 48 PM

TS still complains. I think its because this credentials object is coming from auth0 like we discussed before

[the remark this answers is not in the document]
mhuynh5757: I see, that's strange. Maybe because it's a member of an object, I'm wondering if you just check like this `if (credentials.idToken)` if it would work, but I think it's fine as-is with the `as any` then. LGTM

mhuynh5757 (approved):

PR #32, opened 2023-03-15
Chore/restore package lock

kpchoy:
**Contributing to Twilio**

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -18,6 +18,7 @@
     "@twilio/voice-react-native-sdk": "1.0.0-beta.1",
     "react": "18.1.0",
     "react-native": "0.70.6",
+    "react-native-auth0": "^2.17.1",
     "react-native-safe-area-context": "^4.5.0",
     "react-native-screens": "^3.19.0",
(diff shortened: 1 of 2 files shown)


app/package.json
```diff
@@ -31,7 +31,6 @@
     "@types/jest": "^26.0.23",
     "@types/react": "^18.0.21",
     "@types/react-native": "^0.70.6",
-    "@types/react-native-auth0": "2.14.1",
```
mhuynh5757, line 34: Do we need this package?

mhuynh5757 (changes requested):

app/package.json
kpchoy: I think before I may have added it to extract a class, but I removed that code. The compiler doesn't complain right now.

kpchoy: I don't think we need it will add back later if needed.

mhuynh5757: Sounds good, removing packages always feels good haha.

charliesantos (approved):
👍

PR #33, opened 2023-03-16
feat: secure twiml endpoint with signature validation

kpchoy:
[VBLOCKS-1466](https://issues.corp.twilio.com/browse/VBLOCKS-1466)

Secure TwiML endpoint with signature validation

ref: https://www.twilio.com/blog/how-to-secure-twilio-webhook-urls-in-nodejs

**Contributing to Twilio**

### DIFF
--- a/server/example.env
+++ b/server/example.env
@@ -23,9 +23,6 @@ CALLER_ID=+11234567890
 # Please see https://github.com/twilio/twilio-voice-react-native#incoming-calls
 PUSH_CREDENTIAL_SID=xxxxxxxxxxxxxx
 
-# Twilio TwiML Request Url
-TWIML_REQUEST_URL=http://xxxxxxxxxxxx.ngrok.io/twiml
-
 # The Identifier field on your Auth0 API's settings tab
 # See https://auth0.com/docs/secure/tokens/access-tokens/get-access-tokens
 AUTH0_AUDIENCE=https://xxxxxxxxxx/api
--- a/server/example.env
+++ b/server/example.env
@@ -1,24 +1,35 @@
 # Port number to run the server on
-export PORT=3030
+PORT=3030
 
 # Twilio account sid
-export ACCOUNT_SID=ACxxxxxxxxxxxxxx
+ACCOUNT_SID=ACxxxxxxxxxxxxxx
 
 # Twilio auth token
-export AUTH_TOKEN=xxxxxxxxxxxxxx
+AUTH_TOKEN=xxxxxxxxxxxxxx
 
 # Twilio API key
-export API_KEY_SID=SKxxxxxxxxxxxxxx
+API_KEY_SID=SKxxxxxxxxxxxxxx
 
 # Twilio API secret
-export API_KEY_SECRET=xxxxxxxxxxxxxx
+API_KEY_SECRET=xxxxxxxxxxxxxx
 
 # Twilio TwiML App sid
-export OUTGOING_APPLICATION_SID=APxxxxxxxxxxxxxx
+OUTGOING_APPLICATION_SID=APxxxxxxxxxxxxxx
 
 # Caller ID
-export CALLER_ID=+11234567890
+CALLER_ID=+11234567890
 
 # Push credential for incoming calls
 # Please see https://github.com/twilio/twilio-voice-react-native#incoming-calls
-export PUSH_CREDENTIAL_SID=xxxxxxxxxxxxxx
+PUSH_CREDENTIAL_SID=xxxxxxxxxxxxxx
+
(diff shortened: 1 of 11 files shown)


server/src/routes/twiml.ts
```diff
@@ -1,12 +1,26 @@
...
 export function createTwimlRoute(
-  twilioCredentials: TwilioCredentials,
+  serverConfig: ServerCredentials,
 ): RequestHandler {
   const { VoiceResponse } = twiml;
   return function twimlRoute(req: Request, res: Response) {
+    const twilioSignature = req.headers['x-twilio-signature'] as string;
+    const url = serverConfig.TWIML_REQUEST_URL;
+    const params = req.body;
+    const requestIsValid = validateRequest(
```
mhuynh5757, line 13: I believe there's a function called `validateExpressRequest` or something along those lines in the `twilio` helper library that does all of this work for us, and I think it means we won't need to have `TWIML_REQUEST_URL` as an environment variable for us either, since it will pull the URL info out of the Express request. The latter I'm not too sure on, but the former is definitely worth checking out.

server/yarn.lock
mhuynh5757, line 1: Where did the lockfile changes come from? The `package.json` hasn't changed it seems... are these changes from `main` or something?

mhuynh5757 (changes requested):

server/example.env
```diff
@@ -1,24 +1,35 @@
...
-export CALLER_ID=+11234567890
+CALLER_ID=+11234567890
 
 # Push credential for incoming calls
 # Please see https://github.com/twilio/twilio-voice-react-native#incoming-calls
-export PUSH_CREDENTIAL_SID=xxxxxxxxxxxxxx
+PUSH_CREDENTIAL_SID=xxxxxxxxxxxxxx
+
+# Twilio TwiML Request Url
+TWIML_REQUEST_URL=http://xxxxxxxxxxxx.ngrok.io/twiml
```
charliesantos, line 27: TWIML_REQUEST_URL is usually the customer's domain. This is also called voice request url in your twiml app and should be served via https. Let's update this section to something like the following.
```sh
# Twilio Voice Request URL configured in your TwiML App
VOICE_REQUEST_URL=https://mycompany.com/twiml
```

However, if you can validate the request using what Michael suggested [here](https://github.com/twilio/twilio-voice-react-native-app/pull/33/files#r1139342402), then no need for this change.

charliesantos (changes requested):
Great work @kpchoy . Left one comment.

server/example.env
kpchoy: Yes decided to go with Michael's suggestion, since then we no longer need the TwiML url

server/yarn.lock
kpchoy: I think I either forgot to commit these before, or the merge with main got rid of them. But they are needed for the `oauth-jwt` package

mhuynh5757: We dropped this dependency, no? There should be no difference between the `package.json` files on `main` and `feature/3rd_party_auth` I believe.

mhuynh5757: Ah, nevermind, I see the package. Sorry, remind me what we need that package for haha.

kpchoy: No I added this dependency. I needed to add this to secure the routes with auth0 on the backend.

server/src/__tests__/routes/twiml.test.ts
```diff
@@ -1,9 +1,9 @@
 import { createTwimlRoute } from '../../routes/twiml';
-import { twiml, validateRequest } from 'twilio';
+import { twiml, validateExpressRequest } from 'twilio';
 
 const mockedVoiceResponse = jest.mocked(twiml.VoiceResponse);
 
-jest.mocked(validateRequest);
+jest.mocked(validateExpressRequest);
```
mhuynh5757, line 6: `jest.mocked` is a no-op utility function that just lets you get a `jest.Mock` out of a package, so the use case for this here is something like this

```
const mockedValidateExpressRequest = jest.mocked(validateExpressRequest);
```

and TS will give you things like

```
mockedValidateExpressRequest.mock.calls;
```

without complaints. Since we're not using the return value here, this `jest.mocked` call isn't really doing all that much.

mhuynh5757 (changes requested):
Implementation looks good, just super minor unit test changes recommended.

server/src/__tests__/routes/twiml.test.ts
mhuynh5757, line 1: Do we have a test for the `twiml` route when `validateExpressRequest` returns `false` such that the request is invalid?

mhuynh5757 (approved):

PR #34, opened 2023-03-20
[VBLOCKS-1444] CircleCI config

mhuynh5757:
This PR creates a CircleCI configuration to automatically run some tests for the Reference App and Reference Server.

**Contributing to Twilio**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,7 +14,7 @@ parameters:
 executors:
   default:
     docker:
-      - image: cimg/node:lts
+      - image: cimg/node:16.19
     working_directory: ~/project
 
 ###
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -65,14 +65,14 @@ jobs:
 
   app-test:
     executor: default
-      steps:
-        - attach_project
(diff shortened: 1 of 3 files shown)


charliesantos:
@mhuynh5757 build is failing. Seems to be pretty straightforward fix. Can you please address so we can have a clean build going into main?

.circleci/config.yml
```diff
@@ -0,0 +1,155 @@
...
+    type: boolean
+    default: true # by default pr workflow will get executed.
+
+###
+# Executors
+###
+executors:
+  default:
+    docker:
+      - image: cimg/node:lts
```
charliesantos, line 17: Do we really want to pin on lts? This will break future builds if new lts versions are not compatible with our app. I think we want this to be manual..

charliesantos (changes requested):

mhuynh5757:
I couldn't find a CircleCI image that pinned to `16` (major version) just to latest `16.19` (minor version).

charliesantos (approved):
Great work! It's all green now. I like green builds 😄

PR #38, opened 2023-03-24
[VBLOCKS-1502] Detox pipeline

mhuynh5757:
Add a Detox pipeline to the Reference App

**Contributing to Twilio**


app/index.js
mhuynh5757, line 1: I had to do this to get the bundler detecting the entry point of the project correctly. Unfortunate, but there's not a lot of heavy lifting we do in this file anyways.

mhuynh5757: It should, and Android actually supports this. The issue arises when the bundler is run in `release` mode (again). Android has an option to define a different entry point file in the `app/build.gradle` file, but iOS doesn't seem to have an analogous option that I could find.

mhuynh5757: The bundler is specifically looking for `index.js` when in `release` mode.

kbagchiGWC (commented):
Android changes looks fine.

charliesantos (approved):
Great work @mhuynh5757

charliesantos (approved):

PR #43, opened 2023-03-30
[VBLOCKS-1494] feat: outgoing ux improvements

kpchoy:
[VBLOCKS-1494](https://issues.corp.twilio.com/browse/VBLOCKS-1494)

- don't allow calling empty numbers
- remove back space when no number
- Fix text on toggle to read 'Client' when using number
- auto navigate back to dialer after call has ended (1 second timeout)

Figma Ref: https://www.figma.com/file/IFiFQzWEp1xyM2nuqJCkC2/2022-Q2-React-Voice-SDK?node-id=629-19201&t=uIDqNjfDOVKLVVt8-0

## Before:

dialer

## After:

Android & iOS

Screen Shot 2023-03-30 at 3 46 26 PM

Screen Shot 2023-03-30 at 3 58 49 PM

https://user-images.githubusercontent.com/35968892/228982476-555cbb62-8d2e-47ce-8f82-cd997e0553c0.mov

**Contributing to Twilio**

### DIFF
--- a/app/src/screens/Call/index.tsx
+++ b/app/src/screens/Call/index.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { StyleSheet, View } from 'react-native';
 import Dialpad from '../../components/Dialpad';
 import RemoteParticipant from './RemoteParticipant';
@@ -8,6 +8,9 @@ import SelectAudioOutputDeviceButton from './SelectAudioOutputDeviceButton';
 import ShowDialpadButton from './ShowDialpadButton';
 import HideDialpadButton from './HideDialpadButton';
 import useCall from './hooks';
+import { useNavigation } from '@react-navigation/native';
+import { type StackNavigationProp } from '../../types';
+import { Call as TwilioCall } from '@twilio/voice-react-native-sdk';
 
 const Call: React.FC = () => {
   const {
@@ -54,6 +57,14 @@ const Call: React.FC = () => {
     ),
     [dialpad, hangup, mute, selectAudioOutputDevice],
   );
+
+  const navigation = useNavigation<StackNavigationProp<'App'>>();
+
+  useEffect(() => {
+    if (callStatus === TwilioCall.State.Disconnected) {
+      setTimeout(() => navigation.navigate('Dialer'), 1000);
+    }
+  }, [callStatus, navigation]);
 
   return (
     <View style={styles.container}>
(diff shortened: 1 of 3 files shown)


app/src/screens/Call/index.tsx
```diff
@@ -55,6 +58,14 @@ const Call: React.FC = () => {
     [dialpad, hangup, mute, selectAudioOutputDevice],
   );
 
+  const navigation = useNavigation<StackNavigationProp<'App'>>();
+
+  useEffect(() => {
+    if (callStatus === TwilioCall.State.Disconnected) {
+      setTimeout(() => navigation.navigate('Dialer'), 1000);
```
charliesantos, line 65: Why do you need a setTimeout?

app/src/screens/Dialer/ToggleClientInputButton.tsx
```diff
@@ -17,7 +17,7 @@ const ToggleClientInputButton: React.FC<Props> = ({
   <Button disabled={disabled} size={96} onPress={onPress}>
     <Image source={TypeSource} resizeMode="contain" style={styles.image} />
     <Text style={styles.text}>
-      {recipientType === 'number' ? 'Number' : 'Client'}
+      {recipientType === 'number' ? 'Client' : 'Number'}
```
charliesantos, line 20: I don't get this change. Why switch the two?

app/src/screens/Call/index.tsx
kpchoy: Screen Shot 2023-03-31 at 9 12 18 AM

Without the setTimeout you will not be able to see the "disconnected" UI. I just thought it would be a more user friendly experience. But if you don't think the user needs to see the "disconnected" UI then I will remove the set timeout.

app/src/screens/Dialer/ToggleClientInputButton.tsx
kpchoy: Currently on main we have this: 
dialer

But I believe this is incorrect, since if you are using number dialpad and click the "number" toggle the "client" toggle will be triggered.

This is the figma:
Screen Shot 2023-03-31 at 9 16 52 AM

See how while using number dialpad the "client" toggle is visible.

charliesantos: Ok got it.

charliesantos (approved):

PR #44, opened 2023-04-04
fix(app): dialer hooks refactor

mhuynh5757:
Fix some of the UX of the Dialer screen and refactor the hook logic.

**Contributing to Twilio**

### DIFF
--- a/app/.eslintrc.js
+++ b/app/.eslintrc.js
@@ -7,4 +7,7 @@ module.exports = {
   globals: {
     globalThis: false, // false denotes not-writable
   },
+  rules: {
+    'no-shadow': 'error',
+  },
 };
--- a/app/.eslintrc.js
+++ b/app/.eslintrc.js
@@ -7,7 +7,13 @@ module.exports = {
   globals: {
     globalThis: false, // false denotes not-writable
   },
-  rules: {
-    'no-shadow': 'error',
-  },
+  overrides: [
+    {
+      files: ['*.ts', '*.tsx'],
+      rules: {
+        '@typescript-eslint/no-shadow': ['error'],
+        'no-shadow': 'off',
+      },
+    },
+  ],
 };
--- a/app/.eslintrc.js
+++ b/app/.eslintrc.js
@@ -3,14 +3,8 @@ module.exports = {
   extends: '@react-native-community',
   parser: '@typescript-eslint/parser',
   plugins: ['@typescript-eslint'],
-  overrides: [
-    {
-      files: ['*.ts', '*.tsx'],
-      rules: {
-        '@typescript-eslint/no-shadow': ['error'],
-        'no-shadow': 'off',
-        'no-undef': 'off',
-      },
-    },
-  ],
+  ignorePatterns: ['coverage/'],
+  globals: {
+    globalThis: false, // false denotes not-writable
+  },
 };
--- a/app/src/screens/Dialer/index.tsx
+++ b/app/src/screens/Dialer/index.tsx
@@ -8,52 +8,44 @@ import OutgoingRemoteParticipant from './OutgoingRemoteParticipant';
 import useDialer from './hooks';
 
 const Dialer: React.FC = () => {
-  const {
-    dialpad,
-    isDisabled,
-    outgoingRemoteParticipant,
-    makeOutgoingCall,
-    recipient,
-  } = useDialer();
+  const { dialpad, makeCall, outgoing, recipientToggle } = useDialer();
 
-  const isOutgoingCallButtonDisabled =
-    dialpad.outgoingNumber === '' || isDisabled;
-
-  const isBackSpaceVisible = dialpad.outgoingNumber !== '';
+  const backspaceButton = React.useMemo(
+    () =>
+      dialpad.backspace.isDisabled ? (
+        <View style={styles.emptyButton} />
+      ) : (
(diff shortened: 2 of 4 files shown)


mhuynh5757:
Tested on Android emulator.

mhuynh5757:
@kpchoy can chime in a bit here, but one of the buttons wasn't being enabled/disabled properly. The hook logic refactor isn't strictly necessary but some of the UX logic was in the component instead of the hook and I moved it and saw the opportunity to adjust the hook as a whole. The functional logic is the same, just organized better. There is no JIRA for this ticket, it came out of a conversation between me and Kevin. Probably should have made a ticket to stay on scope but it was a fast fix.

app/.eslintrc.js
```diff
@@ -3,14 +3,8 @@ module.exports = {
   extends: '@react-native-community',
   parser: '@typescript-eslint/parser',
   plugins: ['@typescript-eslint'],
-  overrides: [
-    {
-      files: ['*.ts', '*.tsx'],
-      rules: {
-        '@typescript-eslint/no-shadow': ['error'],
```
charliesantos, line 10: Why remove this? Is this the same as removing the `no-shadow` below?

app/.eslintrc.js
```diff
@@ -3,14 +3,8 @@ module.exports = {
...
-        'no-shadow': 'off',
-        'no-undef': 'off',
```
charliesantos, line 12: Removing this turns on `no-undef` right?

kpchoy:
@charliesantos There is a small bug with this ticket: https://issues.corp.twilio.com/browse/VBLOCKS-1494 . I forgot to test the `client` dialer. And the call button is disabled since I only checked that the `number` field was not empty. The rest is @mhuynh5757 refactoring the hooks. Not all related to the error caused by my ticket.

charliesantos (changes requested):
Code looks good. Just left a couple of comments on eslint rules.
Also, when do you guys think we should start adding tests to our PRs? We have the tooling now right?

app/.eslintrc.js
mhuynh5757: I wasn't sure why these defaults were here to be honest. Removing the whole `overrides` section makes our linting stricter (better, imo).

charliesantos: Ok. Do you know if `no-shadow` rule is enabled by default? I think we need to turn it on if it's disabled after removing this override.

app/src/screens/Dialer/index.tsx
```diff
@@ -8,52 +8,44 @@ import OutgoingRemoteParticipant from './OutgoingRemoteParticipant';
...
-    makeOutgoingCall,
-    recipient,
-  } = useDialer();
+  const { dialpad, makeCall, outgoing, recipientToggle } = useDialer();
 
-  const isOutgoingCallButtonDisabled =
-    dialpad.outgoingNumber === '' || isDisabled;
-
-  const isBackSpaceVisible = dialpad.outgoingNumber !== '';
+  const backspaceButton = React.useMemo(
```
kpchoy, line 13: nit: since this is a component lets capitalize it
```suggestion
  const BackspaceButton = React.useMemo(
```

mhuynh5757: I actually think it looks better capitalized, but I'm not sure if it's technically proper since it's not _really_ a React component, it's an instantiated component and therefore a `JSX.Element`. Minor difference, but there's implications since it's not a "constructor" and it's a "instance" basically. Let me look up the recommended practices.

app/.eslintrc.js
mhuynh5757: It was, but I just pushed an addition that makes it error instead of warn on shadowing.

charliesantos (approved):

app/.eslintrc.js
mhuynh5757: Oops, I think I get (some) of these defaults now. I think that `@typescript-eslint/no-shadow` has different rules than `no-shadow` (which since it's not prefixed I'm assuming is from the base `eslint` rules). So what the RN team did is basically prefer `@typescript-eslint/no-shadow` over `eslint/no-shadow` for `.ts` and `.tsx` files specifically. I think I'm going to revert the `*no-shadow` changes but I'd still like to keep the `no-undef` changes. Sorry for the back-and-forth.

charliesantos: Ok no problem

app/src/screens/Dialer/index.tsx
mhuynh5757: I couldn't find anything concrete on this one, @charliesantos @kpchoy what do you think? I'm open to either, but the current codebase is lowercased.

charliesantos: The convention is not set in stone. But historically, we use `PascalCase` for exported components, `camelCase` otherwise.

kpchoy (approved):

app/src/screens/Dialer/index.tsx
kpchoy: Just thought it looks better, but lets leave it then

charliesantos: I don't have any preference. We can change the convention. But we need to make sure the whole codebase follows for consistency. We cannot change only in one place. If we like `PascalCase` better, please add a ticket to change this on a future sprint :)

PR #45, opened 2023-04-05
[VBLOCKS-1493] - Feature: error handling and logging

kpchoy:
[VBLOCKS-1493](https://issues.corp.twilio.com/browse/VBLOCKS-1493)

SDK Errors and Warnings
- handle error cases

Server 
- try catch implementation for routes
- log request errors

### DIFF
--- a/app/src/store/user.ts
+++ b/app/src/store/user.ts
@@ -89,7 +89,7 @@ export const login = createAsyncThunk<{ accessToken: string; email: string }>(
         email: user.email,
       };
     } catch (error) {
-      return rejectWithValue(error?.error_description);
+      return rejectWithValue(error);
     }
   },
 );
@@ -102,7 +102,7 @@ export const logout = createAsyncThunk<{ accessToken: string; email: string }>(
       await auth0.credentialsManager.clearCredentials();
       return { accessToken: '', email: '' };
     } catch (error) {
-      return rejectWithValue(error?.error_description);
+      return rejectWithValue(error);
     }
   },
 );
--- a/app/src/store/user.ts
+++ b/app/src/store/user.ts
@@ -3,10 +3,18 @@ import config from '../../config';
 import Auth0 from 'react-native-auth0';
 import { State, Dispatch, type AsyncStoreSlice } from './app';
 
-export type UserState = AsyncStoreSlice<{
-  accessToken: string;
-  email: string;
-}>;
+export type UserState = AsyncStoreSlice<
+  {
+    status: 'fulfilled';
+    accessToken: string;
+    email: string;
+  },
+  {
+    status: 'rejected';
+    reason: 'ID_TOKEN_UNDEFINED' | 'LOGIN_ERROR' | 'LOGOUT_ERROR' | undefined;
+    error: any;
+  }
+>;
 
 export const userSlice = createSlice({
   name: 'user',
--- a/app/src/store/user.ts
+++ b/app/src/store/user.ts
@@ -89,7 +89,7 @@ export const login = createAsyncThunk<{ accessToken: string; email: string }>(
         email: user.email,
       };
     } catch (error) {
-      return rejectWithValue('LOGIN_ERROR');
+      return rejectWithValue(error?.error_description);
     }
   },
 );
@@ -102,7 +102,7 @@ export const logout = createAsyncThunk<{ accessToken: string; email: string }>(
       await auth0.credentialsManager.clearCredentials();
       return { accessToken: '', email: '' };
     } catch (error) {
-      return rejectWithValue('LOGOUT_ERROR');
+      return rejectWithValue(error?.error_description);
     }
   },
 );
--- a/app/src/store/user.ts
+++ b/app/src/store/user.ts
@@ -5,12 +5,10 @@ import { State, Dispatch, type AsyncStoreSlice } from './app';
 
 export type UserState = AsyncStoreSlice<
   {
-    status: 'fulfilled';
     accessToken: string;
     email: string;
   },
   {
-    status: 'rejected';
     reason: 'ID_TOKEN_UNDEFINED' | 'LOGIN_ERROR' | 'LOGOUT_ERROR' | undefined;
     error: any;
   }
--- a/app/src/store/user.ts
+++ b/app/src/store/user.ts
@@ -1,7 +1,7 @@
 import { createSlice, createAsyncThunk } from '@reduxjs/toolkit';
 import config from '../../config';
 import Auth0 from 'react-native-auth0';
-import { type AsyncStoreSlice } from './app';
+import { State, Dispatch, type AsyncStoreSlice } from './app';
 
 export type UserState = AsyncStoreSlice<{
   accessToken: string;
@@ -28,7 +28,11 @@ export const userSlice = createSlice({
         return { status: 'fulfilled', ...action.payload };
       })
       .addCase(login.rejected, (_, action) => {
-        return { status: 'rejected', error: action.payload };
+        return {
+          status: 'rejected',
+          reason: action.payload?.reason,
+          error: action.payload?.error || action.error,
+        };
       });
     builder
       .addCase(logout.pending, () => {
@@ -38,7 +42,11 @@ export const userSlice = createSlice({
         return { status: 'fulfilled', ...action.payload };
       })
       .addCase(logout.rejected, (_, action) => {
-        return { status: 'rejected', error: action.payload };
+        return {
+          status: 'rejected',
+          reason: action.payload?.reason,
+          error: action.payload?.error || action.error,
+        };
       });
   },
 });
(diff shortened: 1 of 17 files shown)


charliesantos (changes requested):
Left a few comments. @mhuynh5757 you're more familiar with the RN SDK. Can you please see if there are any errors that we need to handle? Do get some kind of generic error event that we can subscribe to? Do we need to handle any other promise rejections?

@kpchoy are there any warnings that we need to log? Any logic related warnings?
Any debug logs that we need to add? For example, if something goes wrong, do we have enough information in the debug logs to help us troubleshoot with the customer?

In addition to my server related comments, I think we need access logs. Please check this out https://stackoverflow.com/questions/19681226/built-in-access-logs-in-node-js-express-framework
Some of the solutions are outdated, but the idea is the same.

kpchoy:
@charliesantos right now I have subscribed to Connectfailure, Reconnecting, and Disconnected events. https://github.com/twilio/twilio-voice-react-native-app/pull/45/files#diff-7fcac4a84674b05d9e79eab10687887f51e524ebe90006692215a7660d262c19R42

@mhuynh5757 any other ones you suggest?

server/src/routes/twiml.ts
```diff
@@ -1,45 +1,57 @@
...
-    res
-      .header('Content-Type', 'text/xml')
-      .status(200)
-      .send(twimlResponse.toString());
+      res
+        .header('Content-Type', 'text/xml')
+        .status(200)
+        .send(twimlResponse.toString());
+    } catch (e) {
+      error(e);
```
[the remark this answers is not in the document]
kpchoy: I agree this try catch block actually does not make sense where I added it

app/src/store/user.ts
```diff
@@ -102,7 +102,7 @@ export const logout = createAsyncThunk<{ accessToken: string; email: string }>(
       await auth0.credentialsManager.clearCredentials();
       return { accessToken: '', email: '' };
     } catch (error) {
-      return rejectWithValue('LOGOUT_ERROR');
+      return rejectWithValue(error);
```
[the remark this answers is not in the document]
kpchoy: No at the time this is just a subjective message I put for this error. But since auth0 will return an error if any of the async calls fail, I figure it would be best to return their error.

mhuynh5757:
The `voice` object also emits errors. I think it'd be really nice to be able to see if any of our async await calls end up rejecting, it tells us something about the native layer. In effect, if the Android layer wasn't able to construct a CallListenerProxy or whatever for any reason, that rejection could clue us into that. I think the idea of being able to listen to any `voice` rejection (or `call` rejection for that matter) is something we should investigate.

kpchoy:
> In addition to my server related comments, I think we need access logs. Please check this out https://stackoverflow.com/questions/19681226/built-in-access-logs-in-node-js-express-framework Some of the solutions are outdated, but the idea is the same.

@charliesantos I decided to add `winston` https://github.com/winstonjs/winston

We will have 2 log files `combined.log` and `error.log` that customers can send to us. There is some freedom where we can have just 1 or multiple files, also can filter the log level that gets logged to the file.

Screen Shot 2023-04-06 at 4 37 23 PM

This is how the Console will look like: 

Screen Shot 2023-04-06 at 4 36 47 PM

This logging package also comes with a lot of configs, so let me know if you want to change anything @charliesantos @mhuynh5757

kpchoy:
> The `voice` object also emits errors. I think it'd be really nice to be able to see if any of our async await calls end up rejecting, it tells us something about the native layer. In effect, if the Android layer wasn't able to construct a CallListenerProxy or whatever for any reason, that rejection could clue us into that. I think the idea of being able to listen to any `voice` rejection (or `call` rejection for that matter) is something we should investigate.

@mhuynh5757 Just added async error handling for our `/app` let me know what you think of the above commit ^

charliesantos (approved):
:shipit: 👍 
Please make sure build passes (non detox tests)

app/src/store/voice/audioDevices.ts
```diff
@@ -13,17 +13,23 @@ export const getAudioDevices = createAsyncThunk<
   { audioDevices: AudioDeviceInfo[]; selectedDevice: AudioDeviceInfo | null },
   void,
   { state: State; dispatch: Dispatch }
->('voice/getAudioDevices', async () => {
-  const { audioDevices, selectedDevice } = await voice.getAudioDevices();
+>('voice/getAudioDevices', async (_, { rejectWithValue }) => {
+  try {
```
mhuynh5757, line 17: I'm not sure this `try` catch will actually do anything at the moment. I don't believe `voice.getAudioDevices` can fail with how the native code base is currently set up, but I think this is good future-proofing. Something to note, doesn't change the review.

app/src/store/voice/call/activeCall.ts
```diff
@@ -47,8 +47,11 @@ export const disconnectActiveCall = createAsyncThunk<
     if (!call) {
       return rejectWithValue('CALL_UUID_NOT_FOUND');
     }
-
-    await call.disconnect();
+    try {
```
mhuynh5757, line 50: Since we're catching these now, these thunks will never really "fail". So if you try matching fulfillment or rejection using the Redux toolkit matchers, they won't give us what we want.

Logging the error here doesn't also provide too much, the default behavior of an unhandled Promise reject is effectively the same as what we're doing here - arguably more useful. I think we need to use `rejectWithValue` here and have more information in the store about the failure reason. See the above recommendation with having a `reason` string.

server/src/routes/token.ts
```diff
@@ -1,6 +1,7 @@
 import { Request, RequestHandler, Response } from 'express';
 import { jwt } from 'twilio';
 import { TwilioCredentials } from '../common/types';
+import { error } from '../utils/log';
```
mhuynh5757, line 4: Where do we use this? I see an `import` but the GitHub diff doesn't show any uses.

server/src/routes/twiml.ts
```diff
@@ -1,20 +1,28 @@
 import { Request, RequestHandler, Response } from 'express';
 import { twiml, validateExpressRequest } from 'twilio';
 import { ServerCredentials } from '../common/types';
+import { error } from '../utils/log';
 
 export function createTwimlRoute(
   serverConfig: ServerCredentials,
 ): RequestHandler {
   const { VoiceResponse } = twiml;
```
mhuynh5757, line 9: Preference: bind an error handler here.
```
const logError = (msg: string) => {
  error('/twiml ${msg}`);
};
```
Later you can do something like
```
const msg = 'Missing "To".';
logError(msg);
res.status(400).send(msg);
```
So we're less prone to having different return values and error logs.

server/src/utils/log.ts
```diff
@@ -1,3 +1,49 @@
...
+      ),
+    }),
+  );
+}
+
+export const error = (message: string) =>
+  logger.error({
+    level: 'error',
+    message,
+  });
```
mhuynh5757, line 40:
```suggestion
  });

```

server/src/utils/log.ts
mhuynh5757, line 1: Great work with using `winston`, really like the idea of using a standard lib for this!

app/src/store/voice/audioDevices.ts
```diff
@@ -61,8 +67,8 @@ export const audioDevicesSlice = createSlice({
       .addCase(getAudioDevices.fulfilled, (_, action) => {
         return { status: 'fulfilled', ...action.payload };
       })
-      .addCase(getAudioDevices.rejected, () => {
-        return { status: 'rejected' };
+      .addCase(getAudioDevices.rejected, (_, action) => {
+        return { status: 'rejected', error: action.payload };
```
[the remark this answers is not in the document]
kpchoy: I dont really get how to implement the reason into the `addCase`.

```
if the promise resolved with a rejectWithValue(value) return value, dispatch the rejected action with the value passed into action.payload and 'Rejected' as action.error.message
```

Did you want? :
```
return { status: 'rejected', reason: action.payload.reason, error: action.payload.error };
```

app/src/store/user.ts
```diff
@@ -1,7 +1,7 @@
 import { createSlice, createAsyncThunk } from '@reduxjs/toolkit';
 import config from '../../config';
 import Auth0 from 'react-native-auth0';
-import { type AsyncStoreSlice } from './app';
+import { State, Dispatch, type AsyncStoreSlice } from './app';
 
 export type UserState = AsyncStoreSlice<{
```
mhuynh5757, line 6: Let's add the rejection info to the state.
```
AsyncStoreSlice<fulfilled type, rejected type>;
```

mhuynh5757 (approved):

mhuynh5757 (changes requested):
Sorry, let's address that last comment first.

app/src/store/user.ts
```diff
@@ -3,10 +3,18 @@ import config from '../../config';
 import Auth0 from 'react-native-auth0';
 import { State, Dispatch, type AsyncStoreSlice } from './app';
 
-export type UserState = AsyncStoreSlice<{
-  accessToken: string;
-  email: string;
-}>;
+export type UserState = AsyncStoreSlice<
+  {
+    status: 'fulfilled';
```
mhuynh5757, line 8: Sorry, I know you're probably sick of having to change things.

The types we pass to `AsyncStoreSlice` don't need `status` defined on them since they get merged by the helper type itself. So if you do something like `AsyncStoreSlice<T>`, then `T` gets merged automatically with `{ status: 'fulfilled' }`.

mhuynh5757 (changes requested):

mhuynh5757 (approved):

PR #49, opened 2023-04-19
[VBLOCKS-1352] Incoming call registration

mhuynh5757:
This PR introduces new application bootstrapping actions. Within the bootstrapping action, Twilio Voice access token and registration occurs.

**Contributing to Twilio**

### DIFF
--- a/app/.gitignore
+++ b/app/.gitignore
@@ -65,3 +65,9 @@ buck-out/
 
 # Jest
 coverage/
+
+###
+# Secrets
+###
+# Google Services
+google-services.json
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -1,11 +1,11 @@
 import { NavigationContainer } from '@react-navigation/native';
 import React from 'react';
 import { Provider } from 'react-redux';
-import { createStore } from './store/app';
+import { createStore, logActionType } from './store/app';
 import StackNavigator from './screens/StackNavigator';
 import { bootstrapApp } from './store/bootstrap';
 
-const store = createStore();
+const store = createStore(logActionType);
 
 const App = () => {
   React.useEffect(() => store.dispatch(bootstrapApp()).abort, []);
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -8,6 +8,17 @@ import { bootstrapApp } from './store/bootstrap';
 const store = createStore(logActionType);
 
 const App = () => {
+  /**
+   * NOTE:
+   * When Redux Toolkit dispatches a Thunk, it will put an `AbortController`
+   * into the returned Promise as the `abort` member. Invoking the `abort`
+   * method will prevent further actions from being dispatched in the Thunk.
+   *
+   * When a React component is unmounted, it will invoke the return value of any
+   * `useEffect` functions. In this case, if the `App` component is unmounted
+   * then the `abort` function is called so the `bootstrapApp` action can no
+   * longer dispatch actions.
+   */
   React.useEffect(() => store.dispatch(bootstrapApp()).abort, []);
 
   return (
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -1,9 +1,11 @@
 import { NavigationContainer } from '@react-navigation/native';
 import React from 'react';
 import { Provider } from 'react-redux';
-import { store } from './store/app';
+import { createStore } from './store/app';
 import StackNavigator from './screens/StackNavigator';
 import { bootstrapApp } from './store/bootstrap';
+
+const store = createStore();
 
 const App = () => {
   React.useEffect(() => {
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -8,9 +8,7 @@ import { bootstrapApp } from './store/bootstrap';
 const store = createStore();
 
 const App = () => {
-  React.useEffect(() => {
-    store.dispatch(bootstrapApp());
-  }, []);
+  React.useEffect(() => store.dispatch(bootstrapApp()).abort, []);
 
   return (
     <Provider store={store}>
--- a/app/src/App.tsx
+++ b/app/src/App.tsx
@@ -3,8 +3,13 @@ import React from 'react';
 import { Provider } from 'react-redux';
 import { store } from './store/app';
 import StackNavigator from './screens/StackNavigator';
+import { bootstrapApp } from './store/bootstrap';
 
 const App = () => {
+  React.useEffect(() => {
+    store.dispatch(bootstrapApp());
+  }, []);
+
   return (
     <Provider store={store}>
       <NavigationContainer>
--- a/app/src/screens/Dialer/hooks.ts
+++ b/app/src/screens/Dialer/hooks.ts
@@ -119,11 +119,7 @@ const useMakeOutgoingCall = (
       console.error(callAction.payload || callAction.error);
     }
 
-    if (makeOutgoingCallAction.fulfilled.match(callAction)) {
-      navigation.navigate('Call');
-    } else {
-      console.error(callAction.payload || callAction.error);
-    }
+    navigation.navigate('Call');
   }, [dispatch, navigation, recipientType, to]);
 
   const isDisabled = React.useMemo(() => {
--- a/app/src/screens/Dialer/hooks.ts
+++ b/app/src/screens/Dialer/hooks.ts
@@ -6,7 +6,7 @@ import {
   makeOutgoingCall as makeOutgoingCallAction,
   type RecipientType,
 } from '../../store/voice/call/outgoingCall';
-import { getToken } from '../../store/voice/token';
+import { getAccessToken } from '../../store/voice/accessToken';
 import { type StackNavigationProp } from '../../types';
 import { useActiveCall } from '../../hooks/activeCall';
 
@@ -103,8 +103,8 @@ const useMakeOutgoingCall = (
   const to = recipientType === 'client' ? outgoingClient : outgoingNumber;
 
   const handle = React.useCallback(async () => {
-    const tokenAction = await dispatch(getToken());
-    if (getToken.rejected.match(tokenAction)) {
+    const tokenAction = await dispatch(getAccessToken());
+    if (getAccessToken.rejected.match(tokenAction)) {
       console.error(tokenAction.payload || tokenAction.error);
       return;
     }
--- a/app/src/screens/Dialer/hooks.ts
+++ b/app/src/screens/Dialer/hooks.ts
@@ -117,6 +117,7 @@ const useMakeOutgoingCall = (
     );
     if (makeOutgoingCallAction.rejected.match(callAction)) {
       console.error(callAction.payload || callAction.error);
+      return;
     }
 
     navigation.navigate('Call');
--- a/app/src/screens/SignIn/index.tsx
+++ b/app/src/screens/SignIn/index.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { StyleSheet, View, Image, Text, TouchableOpacity } from 'react-native';
 import { useDispatch } from 'react-redux';
-import { login } from '../../store/user';
+import { loginAndRegister } from '../../store/voice/registration';
 import { type Dispatch } from '../../store/app';
 
 const ArrowForward = require('../../../assets/icons/arrow-forward.png');
@@ -67,8 +67,8 @@ const SignIn: React.FC = () => {
   const dispatch = useDispatch<Dispatch>();
 
   const handleLogin = async () => {
-    const loginAction = await dispatch(login());
-    if (login.rejected.match(loginAction)) {
+    const loginAction = await dispatch(loginAndRegister());
+    if (loginAndRegister.rejected.match(loginAction)) {
       console.error(loginAction.payload || loginAction.error);
     }
   };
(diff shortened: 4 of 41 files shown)


app/src/store/bootstrap.ts
```diff
@@ -0,0 +1,49 @@
...
+  | {
+      reason: 'GET_ACCESS_TOKEN_REJECTED';
+    }
+  | {
+      reason: 'REGISTER_REJECTED';
+    };
+
+type BootstrapAppFulfillValue = 'NOT_LOGGED_IN' | 'LOGGED_IN';
+
+export const bootstrapApp = createAsyncThunk<
```
kpchoy, line 20: Why do we need to "bootstrapApp" now? What was wrong with the previous login implementation?

app/src/store/voice/accessToken.ts
```diff
@@ -0,0 +1,125 @@
...
+      error: tokenTextResult.reason,
+    });
+  }
+
+  const token = tokenTextResult.value;
+  return token;
+});
+
+/**
+ * Slice
```
kpchoy, line 77: I don't think this "slice" comment is necessary since line 79 says `AsyncStoreSlice`

app/src/store/voice/call/outgoingCall.ts
```diff
@@ -49,9 +49,11 @@ export const makeOutgoingCall = createAsyncThunk<
       outgoingCall.on(TwilioCall.Event.Reconnecting, (error) =>
         console.error('Reconnecting:', error),
```
kpchoy, line 50: Should we add `if (error)` to these cases as well?

server/src/routes/twiml.ts
```diff
@@ -1,31 +1,31 @@
...
+    log(`/twiml ${msg}`);
   };
 
   return function twimlRoute(req: Request, res: Response) {
     const requestIsValid = validateExpressRequest(req, serverConfig.AUTH_TOKEN);
 
     if (!requestIsValid) {
       const msg = 'Unauthorized Twilio signature';
-      logError(msg);
+      logMsg(msg);
```
kpchoy, line 20: Why log instead of error these messages? Aren't these all failed requests?

kpchoy (changes requested):
Can you make sure the Android Detox CI passes before merging

charliesantos:
Great stuff @mhuynh5757 . I like the improvements. These will definitely help us and our customers in the future. Arguably, some of the changes are probably not required for the completion of the story. As a note for the future, let's draw the line between what is required now and what can be pushed out for later to prevent unnecessary delays and additional work for the team. Me and @kpchoy also needs to spend some time to review 😉. But overall, good work! I will do a deeper review and provide feedback. Just need more time to finish. Also, let's merge @kpchoy 's PR first and make sure your PR is green.

app/src/screens/SignIn/index.tsx
```diff
@@ -66,13 +66,9 @@ const styles = StyleSheet.create({
 const SignIn: React.FC = () => {
   const dispatch = useDispatch<Dispatch>();
 
-  React.useEffect(() => {
-    dispatch(checkLoginStatus());
```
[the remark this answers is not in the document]
charliesantos: +1 to @kpchoy 's comment.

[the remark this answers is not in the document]
mhuynh5757: Yep, still works.

app/src/screens/StackNavigator.tsx
```diff
@@ -11,25 +12,34 @@ const Stack = createNativeStackNavigator<StackParamList>();
...
-      )}
-    </Stack.Navigator>
-  );
+
+  if (user === null) {
+    return <Text>Application not bootstrapped.</Text>;
+  }
+
+  if (user?.status === 'pending') {
+    return null;
```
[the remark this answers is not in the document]
mhuynh5757: Good point, but I think it should be part of the UI/UX tickets. This is temporary for now.

app/src/store/bootstrap.ts
mhuynh5757: Having a bootstrapping action allows us to place more complex logic into our app and have it start somewhere outside of a component. I issue this action in the `App.tsx` file before even rendering the app, and so by the time the rendering is done we can already skip the login screen if the user was already logged in.

app/src/store/voice/call/outgoingCall.ts
mhuynh5757: These events should only ever be emitted when there are errors. Meaning that those errors are never `undefined` and are always truthy. The only exception is the `Disconnected` event where the listener param is `error?` since a disconnect can occur with or without an error.

server/src/routes/twiml.ts
mhuynh5757: This is a good question, I would say errors should be emitted if they are the fault of the server code, but in this case it's the fault of the request. What do you think? cc @charliesantos

app/src/screens/StackNavigator.tsx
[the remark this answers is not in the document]
charliesantos: Which ticket is this part of? Let's make sure it's captured or else we'll forget this.

app/src/store/__tests__/registration.test.ts
```diff
@@ -0,0 +1,140 @@
...
+  })),
+}));
+
+describe('registration', () => {
+  let store: Store;
+  const dispatchedActions: any[] = [];
+
+  beforeEach(() => {
+    dispatchedActions.splice(0);
+    const logAction: Middleware = () => (next) => (action) => {
```
charliesantos, line 24: Ha, so this is the middleware we talked about? I don't have preference between this and jest mocks. I'll leave it up to you.

app/src/store/bootstrap.ts
charliesantos: The previous login implementation is also good, but I think this is a good improvement.

server/src/routes/twiml.ts
charliesantos: We need to add loglevels and define what each level means e.g. info, warn, debug, error. For now, let us use `log` for consistency.

server/src/utils/auth.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export type UserInfoResponse =
+  | {
+      success: true;
+      userInfo: UserInfo;
+    }
+  | {
+      success: false;
+      reason: 'AXIOS_ERROR';
+      error: AxiosError;
```
charliesantos, line 21: Don't make this specific to axios. We're basically exposing what library we're using for API calls. This should be an error that we define. Maybe `AUTH_ERROR`?

server/src/utils/auth.ts
```diff
@@ -0,0 +1,46 @@
...
+  const headers = { 'Authorization': `Bearer ${auth0Token}` };
+  const response = await axios.get(userInfoUrl, { headers })
+    .then((value) => ({ success: true as const, value }))
+    .catch((error) => ({ success: false as const, error }));
+
+  if (!response.success) {
+    return {
+      success: false,
+      reason: 'AXIOS_ERROR',
+      error: response.error,
```
charliesantos, line 38: What is inside `error` object? This is coming from axios right? Can we extract only the information that is needed?

charliesantos (changes requested):

app/src/screens/StackNavigator.tsx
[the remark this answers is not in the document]
mhuynh5757: Added a note to this ticket https://issues.corp.twilio.com/browse/VBLOCKS-1349

app/.gitignore
```diff
@@ -65,3 +65,9 @@ buck-out/
 
 # Jest
 coverage/
+
+###
+# Secrets
+###
+# Google Services
+google-services.json
```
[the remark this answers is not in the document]
mhuynh5757: Yes, it's for building the app. The Android SDK uses Firebase Cloud Messaging for push notifications and this file is required. It's generated when making a Firebase Android App as part of setting up a Firebase project.

app/src/App.tsx
```diff
@@ -1,10 +1,15 @@
...
 import { Provider } from 'react-redux';
-import { store } from './store/app';
+import { createStore, logActionType } from './store/app';
 import StackNavigator from './screens/StackNavigator';
+import { bootstrapApp } from './store/bootstrap';
+
+const store = createStore(logActionType);
 
 const App = () => {
+  React.useEffect(() => store.dispatch(bootstrapApp()).abort, []);
```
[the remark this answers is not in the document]
mhuynh5757: Redux toolkit async action creators automatically put an `AbortController` in the returned `Promise` when dispatching them. Reading up on the Redux toolkit docs recommends this so the `bootstrapApp` action stops dispatching further actions if `AbortController` is used to abort the thunk. The return value of `useEffect` is invoked when the component is dismounted, which in this case should really never happen since it's the root component but just in case.

app/src/screens/Dialer/hooks.ts
```diff
@@ -119,11 +119,7 @@ const useMakeOutgoingCall = (
       console.error(callAction.payload || callAction.error);
     }
 
-    if (makeOutgoingCallAction.fulfilled.match(callAction)) {
-      navigation.navigate('Call');
-    } else {
-      console.error(callAction.payload || callAction.error);
-    }
+    navigation.navigate('Call');
```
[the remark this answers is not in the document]
mhuynh5757: These are duplicate lines, there's some above logic that returns on dispatch-rejection.

app/src/store/__tests__/registration.test.ts
mhuynh5757: Yep, this is it. Super simple so I'm not _too_ concerned about complicating actual app logic, and it helps me test tremendously.

app/src/store/__tests__/store.test.ts
```diff
@@ -33,14 +34,19 @@ jest.mock('@twilio/voice-react-native-sdk', () => {
   return { Call: MockCall };
 });
 
-it('works good', async () => {
+it('works', async () => {
```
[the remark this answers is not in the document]
mhuynh5757: Haha, yeah, I keep forgetting about this one. Will change.

app/src/store/app.ts
```diff
@@ -1,16 +1,32 @@
...
 import { voiceReducer } from './voice';
 
-export const store = configureStore({
-  reducer: {
-    voice: voiceReducer,
-  },
-});
+export const logActionType: Middleware = () => (next) => (action) => {
+  console.log(
+    action.type.match(/\/rejected$/g)
```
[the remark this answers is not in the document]
mhuynh5757: We will still be able to differentiate since the actual `action.type` value is also logged. For example, the logs will look something like this:
```
voice/getAccessToken/pending
voice/getAccessToken/rejected SOME_ERROR_PAYLOAD
```

kpchoy (approved):

app/src/store/user.ts
```diff
@@ -86,7 +107,7 @@ export const login = createAsyncThunk<
   {
     state: State;
     dispatch: Dispatch;
-    rejectValue: { reason: 'ID_TOKEN_UNDEFINED' | 'LOGIN_ERROR'; error: any };
+    rejectValue: { reason: 'ID_TOKEN_UNDEFINED' | 'LOGIN_ERROR'; error?: any };
```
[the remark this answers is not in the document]
mhuynh5757: Good point, should be never. But as I was looking at all of our async action creators, I wanted them all to follow a similar pattern as such:

```ts
type RejectValue =
  | {
      reason: 'foo';
      error: SerializedError;
    }
  | {
      reason: 'bar';
    };
```

And I wanted to change all of our async creators to this pattern, but I think it should go into another ticket/PR since this one is so bloated already.

So making the error optional here for the `ID_TOKEN_UNDEFINED` reason is the smallest change I could think of that still is technically correct just for the sake of this PR. Ideally, the ticket I just mentioned is to clean up all the action creators so they follow the same pattern.

server/src/utils/auth.ts
mhuynh5757: To be honest, I looked at the Auth0 API docs and they're not very clear about the things we can expect from their API, error or success. The payloads they return are not clearly documented.

mhuynh5757: I'm probably not looking in the right place, though.

app/.gitignore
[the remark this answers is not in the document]
mhuynh5757: Here's a ticket for it https://issues.corp.twilio.com/browse/VBLOCKS-1631

app/src/App.tsx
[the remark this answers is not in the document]
charliesantos: Can you put a comment above this line that explains it? It's not immediately clear.

server/src/utils/auth.ts
charliesantos: Can you try triggering an error and log it to see how it looks? We'll be able to see the error format that way and only grab the info we need, not the whole thing.

app/src/store/__tests__/store.test.ts
```diff
@@ -34,7 +34,7 @@ jest.mock('@twilio/voice-react-native-sdk', () => {
   return { Call: MockCall };
 });
 
-it('works', async () => {
+it('can make an outgoing call and mute it', async () => {
```
charliesantos, line 37: >it('should make...

server/src/utils/auth.ts
charliesantos: I see you addressed this. Resolving.

charliesantos (changes requested):
I left a few more responses @mhuynh5757 .

charliesantos (approved):

PR #52, opened 2023-05-03
fix(server): types and ci typecheck

mhuynh5757:
This PR fixes some broken types around server auth and adds a typecheck step in the CI build.

**Contributing to Twilio**

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -226,15 +226,15 @@ jobs:
           root: .
           paths: .
 
-  server-typecheck:
+  server-build:
     executor: default
     steps:
       - attach_project
       - run:
-          name: "Server: typecheck"
+          name: "Server: build"
           command: |
             cd server
-            yarn typecheck
+            yarn build
             cd ..
 
   server-test:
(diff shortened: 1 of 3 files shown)


kpchoy (approved):

charliesantos (changes requested):
Can you include the build command to make sure they don't fail?

mhuynh5757:
The build command does the same thing as the typecheck command, but it generates artifacts whereas typecheck won't. Now that I think about it, we don't need typecheck if you want to remove it and just do build.

charliesantos:
> The build command does the same thing as the typecheck command, but it generates artifacts whereas typecheck won't. Now that I think about it, we don't need typecheck if you want to remove it and just do build.

I understand. That's why we should do build because we might put more scripts as part of build.

mhuynh5757:
I see, I'm fairly certain since
```
yarn build === tsc
yarn typecheck === tsc --noEmit
```
but better safe than sorry. I'll just do a `build` and get rid of the `typecheck`

charliesantos (approved):

PR #53, opened 2023-05-03
[VBLOCKS-1361] feat(app): incoming call screen

mhuynh5757:
This PR implements the components for the Incoming call screen, and moves some reusable components to the appropriate places.

Please see the attached figures for how the components look. Please keep in mind that they are not hooked up to any app state and are not completely functional.

**Contributing to Twilio**

Figure 1:

The new loading screen. The spinner will only show after 500ms of loading.

Figure 2:

The incoming call screen.

Figure 3:

The incoming call notification on top of the dialer screen.


app/assets/icons/twilio-logo.png
mhuynh5757, line 1: I replaced the existing image with one from Kara, it's higher definition.

app/src/components/Call/EndCallButton.tsx
mhuynh5757, line 1: Moved these files to the "reusable components" folder. Let me know if the names should be changed, they simple reflect the name of the icon file.

app/src/screens/IncomingCall/hooks.ts
mhuynh5757, line 1: Temporary hooks, will be elaborated and hooked up to app state with a future ticket.

mhuynh5757:
The existing functionality is still intact, tested thoroughly with Android emulator.

app/assets/icons/twilio-logo.png
kpchoy: I got it from the figma too, but I also noticed the quality wasn't that good. How did you get a better quality image?

app/src/screens/IncomingCall/hooks.ts
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { type Props as RemoteParticipantProps } from '../../components/Call/RemoteParticipant';
+
+export const useIncomingRemoteParticipant = () => {
+  const [title] = React.useState<string>('foo');
```
kpchoy, line 5: Why not just set this to a const? Since you don't `setState`

kpchoy: The function seems unnecessary?

app/src/screens/StackNavigator.tsx
```diff
@@ -18,7 +20,7 @@ const StackNavigator = () => {
   }
 
   if (user?.status === 'pending') {
-    return null;
+    return <LoadingScreen />;
```
kpchoy, line 23: nice 👍

app/src/screens/IncomingCall/hooks.ts
mhuynh5757: It's temporary, will be filled out with this ticket https://issues.corp.twilio.com/browse/VBLOCKS-1379

charliesantos:
@mhuynh5757 great looking screenshots! Few questions:
- Do we have screenshots for incoming call notifications? I understand it uses the default notification in the RN SDK, but it should still work right?
- Probably not in this PR, but will we be getting calls even if the app is not in the foreground?
- Probably not in this PR as well, do we hide incoming notifications if the app is in the foreground?

app/src/components/Call/EndCallButton.tsx
```diff
@@ -7,7 +7,11 @@ const HangupSource = require('../../../assets/icons/end-call.png');
 export type Props = Pick<ButtonProps, 'disabled' | 'onPress'>;
 
 const HangupButton: React.FC<Props> = ({ disabled, onPress }) => (
-  <Button size={96} disabled={disabled} onPress={onPress}>
+  <Button
+    accessibilityLabel="end call"
```
charliesantos, line 11: Should we capitalize?

app/src/components/Call/MakeCallButton.tsx
```diff
@@ -7,7 +7,11 @@ const MakeOutgoingCallSource = require('../../../assets/icons/make-call.png');
 export type Props = Pick<ButtonProps, 'disabled' | 'onPress'>;
 
 const MakeOutgoingCallButton: React.FC<Props> = ({ disabled, onPress }) => (
-  <Button disabled={disabled} size={96} onPress={onPress}>
+  <Button
+    accessibilityLabel="make call"
```
charliesantos, line 11: Same here, should we capitalize?

app/src/components/Call/RemoteParticipant.tsx
```diff
@@ -5,7 +5,7 @@ export type Props = { title: string; subtitle: string };
 
 const RemoteParticipant: React.FC<Props> = ({ title, subtitle }) => {
   return (
-    <View style={styles.container}>
+    <View style={styles.container} accessibilityLabel="remote participant">
```
charliesantos, line 8: should we capitalize?

app/src/components/LoadingScreen.tsx
```diff
@@ -0,0 +1,36 @@
...
+  container: {
+    alignItems: 'center',
+    display: 'flex',
+    flexDirection: 'column',
+    height: '100%',
+    justifyContent: 'center',
+  },
+});
+
+export default LoadingScreen;
```
charliesantos, line 36: This is called `LoadingScreen` but it's not under the screen folder. Any reasons why? Also, instead of calling it `LoadingScreen`, can we remove the `Screen` part to be consistent with other naming? For example, we don't call `Home` screen `HomeScreen` :)

app/src/screens/StackNavigator.tsx
charliesantos: +1!

charliesantos (commented):
It looks really good @mhuynh5757 . Great work. I left only a few comments. But it looks good overall.

app/src/components/LoadingScreen.tsx
mhuynh5757: Good question. It's not a screen at all, maybe just `Loading` then?

app/src/components/Call/RemoteParticipant.tsx
mhuynh5757: I wasn't sure actually, accessibility is something I'm not familiar with at all. Is the standard to capitalize every word?

app/src/components/LoadingScreen.tsx
charliesantos: How about `Busy` or `Spinner`? I prefer `Busy`. But if you have strong opinion on `Loading`, then we can go with that.

app/src/components/Call/RemoteParticipant.tsx
charliesantos: Not sure either. Can we remove it? I think accessibility is a big effort in general.

mhuynh5757: I agree, it's a huge effort. Unfortunately, I don't think we can do any component testing without at least a little bit of labeling or hinting.

charliesantos: Gotcha. We can leave it as is for now 👍

charliesantos (approved):
The notification does not look to bad 😄 
Great work Michael 👍

charliesantos (approved):

kpchoy (approved):

PR #55, opened 2023-05-04
[VBLOCKS-1673] : App ENV file

kpchoy:
[VBLOCKS-1673](https://issues.corp.twilio.com/browse/VBLOCKS-1673)
**Contributing to Twilio**

- add package react-native-config
- env file for app
- no longer need to configure build.gradle


kpchoy:
Since my task involves two parts:
1. App Env File
2. e2e test
Lets do 2 PRs to keep it simple. This first PR just involves the APP .env file upgrade.

.circleci/config.yml
```diff
@@ -35,6 +35,15 @@ commands:
   ###
   # App commands
   ###
+  app-save-env-file:
+    steps:
+      - run:
+          name: "App: save env file"
+          command: |
+            cd app 
+            echo $APP_ENV | base64 -d > .env
```
charliesantos, line 44: Is `APP_ENV` a CircleCI environment variable?

app/README.md
```diff
@@ -8,59 +8,64 @@ Find the `Domain` and `Client ID` found in the `Settings` tab of your Auth0 Appl
 
 ![auth0nativesettings](https://user-images.githubusercontent.com/35968892/226760563-fbc6e3cf-2d63-4fa1-8fa0-eae98dd80859.png)
 
-Add the Application `Domain` and `Client ID` to the app config file
+Add the Application `Domain` and `Client ID` to the app env file
```
charliesantos, line 11: For all the steps requiring to add something to the .env file, I think we should delete them, and just add them in the example.env template. Then in the template, that's where we will explain what the environment variable is for. That way, all environment variable documentations are all in just one place, and not all over the READMEs. Additionally, environment variables should be `SCREAMING_SNAKE_CASE_LIKE_THIS` 😄  (i.e. all caps snake case)

charliesantos (changes requested):
Great! I left a couple of comments.

.circleci/config.yml
kpchoy: Yes I set this one up similar to how @mhuynh5757 setup Google Services JSON in line 62

app/README.md
kpchoy: I like that, it would be a lot more clean. Are we okay with deleting the screen shots then?

charliesantos: Yes. As long as we describe it properly in the example.env files.

.circleci/config.yml
charliesantos: Gotcha. I think this is a different situation where Google Services JSON is generated from google. But the .env file is generated by us. Instead of one single APP_ENV in circle ci, we should add each of the environment variables in circle ci instead. With that, you probably won't need to modify this config.yml

app/README.md
```diff
@@ -4,63 +4,50 @@ The Twilio Voice React Native Reference App is an example implementation of the
...
 for your respective target platform.
 
 ### Setting it up with the Server
 
-Once the Reference Server is up and running, the Reference App can be directed to it by modifying the following line in `app/src/util/fetch.ts`.
+Once the Reference Server is up and running, the Reference App can be directed to it by modifying the following line in `.env`.
+
 ```
-const defaultUrl = ...;
+DEFAULT_URL = ...;
```
mhuynh5757, line 50:
```suggestion
DEFAULT_URL=...;
```

app/src/store/user.ts
```diff
@@ -67,8 +67,8 @@ export const userSlice = createSlice({
 });
 
 const auth0 = new Auth0({
-  domain: config.domain,
-  clientId: config.clientId,
+  domain: Config.DOMAIN_NAME!,
```
mhuynh5757, line 70: Should we check if these are defined or not, instead of using the post-fix `!` operator?

app/package.json
```diff
@@ -21,6 +21,7 @@
     "react": "18.1.0",
     "react-native": "0.70.6",
     "react-native-auth0": "^2.17.1",
+    "react-native-config": "^1.5.0",
```
mhuynh5757, line 24: I took a look at the package on `npm`, it's fairly popular at `200k` DLs per week. I think that should be OK? What do you guys think? I know that DLs per week isn't really a good metric of how well supported a package will be in the future, but it sort of works.

charliesantos: I think it's ok if it satisfies our need.

app/src/store/user.ts
charliesantos: These should required.

kpchoy: Screen Shot 2023-05-11 at 10 08 57 AM
Without the `!` I get a type error. 
The type is defined in the package:
Screen Shot 2023-05-11 at 10 13 03 AM
Any suggestions?

mhuynh5757: Let's try a type-guard before we use it. We should probably make some `util/env` module that does this checking for us.
```ts
if ('domain' in Config && typeof Config.domain !== 'undefined') {
  ...
}
```
Should make the compiler happy.

charliesantos: I'm all for checking things like this but for this situation, what is the benefit other than making the compiler happy?
We know that these values are required but adding checks like this assumes they can be optional.

charliesantos: I think the usage of `!` operator in this case is appropriate. It tells the compiler that this value cannot be undefined or null and trust that it should be there. I believe this is the definition of this operator correct?

charliesantos: @kpchoy , do we get an error if the environment variables are not set? Does it fail silently or do we get a hint that the values are missing? If it fails silently, I'm +1 on Michael's suggestion, but we need to throw an error letting the dev know that the value is missing.

kpchoy: Yes we do get a warning if env variables are not set. If I dont include the CLIENT_ID, this is the error:

Screen Shot 2023-05-11 at 12 03 38 PM

mhuynh5757: Which module is throwing this error? Is it getting all the way to the Auth0 module first, and then once Auth0 recognizes that the env var wasn't passed, it throws? My thought process is that our env helper module would parse the environment for all the required variables and then throw there instead of relying on dependencies to throw, sort of how the Reference Server is doing it.

charliesantos: > Which module is throwing this error? Is it getting all the way to the Auth0 module first, and then once Auth0 recognizes that the env var wasn't passed, it throws? My thought process is that our env helper module would parse the environment for all the required variables and then throw there instead of relying on dependencies to throw, sort of how the Reference Server is doing it.

Thanks for providing more details @mhuynh5757 . I like this better.

kpchoy: I have implemented the helper idea 👍

.circleci/config.yml
```diff
@@ -35,6 +35,19 @@ commands:
...
+          name: "App: save env file"
+          command: |
+            cd app 
+            echo "DOMAIN_NAME=$DOMAIN_NAME" >> .env
+            echo "CLIENT_ID=$CLIENT_ID" >> .env
+            echo "AUTH0_SCOPE=$AUTH0_SCOPE" >> .env
+            echo "AUDIENCE=$AUDIENCE" >> .env
+            echo "DEFAULT_URL=$DEFAULT_URL" >> .env
+            cd ..
+
```
charliesantos, line 50: Are these the only environment variables needed? Do you need to add vblocks context somewhere in this config.yml file?

app/example.env
```diff
@@ -0,0 +1,22 @@
+# Auth0 related configurations
```
charliesantos, line 1: Maybe add something like the following at the top of this file
`# Add your environment variables to this file and rename it to .env`

app/src/util/env.ts
```diff
@@ -0,0 +1,10 @@
+import Config from 'react-native-config';
+
+export const getEnvVariable = (envVar: string): string => {
+  if (envVar in Config && typeof Config[envVar] !== 'undefined') {
+    return Config[envVar] as string;
+  } else {
+    console.error(`Missing ${envVar} in .env`);
```
charliesantos, line 7: Thoughts on throwing an error instead? This is fatal correct?
```ts
throw new Error(`Missing ${envVar} in .env`);
```

charliesantos (commented):
Looks great Kevin. Just left a few minor comments.

.circleci/config.yml
kpchoy: These are the only ones needed for the App. We will later need to add vblocks context for the server, but that will be in `server/env`

app/src/util/env.ts
kpchoy: Throwing an error is better 👍 . Without an envVar the app will crash or not work anyways

charliesantos (approved):

app/example.env
mhuynh5757: Nit: prefer to say something along the lines of "make a copy of this file with the name .env and populate the file with your configuration". Keyword here being "copy and name" instead of "rename".

.circleci/config.yml
mhuynh5757: Now that I'm looking at this again, is this necessary. We're copying env vars to a file, which is the fallback when the env vars are not present. Is that the behavior of the module? That it falls back to the `.env` file for env vars that aren't present?

mhuynh5757 (changes requested):

app/README.md
```diff
@@ -4,63 +4,50 @@ The Twilio Voice React Native Reference App is an example implementation of the
...
 
-Once the Reference Server is up and running, the Reference App can be directed to it by modifying the following line in `app/src/util/fetch.ts`.
+Once the Reference Server is up and running, the Reference App can be directed to it by modifying the following line in `.env`.
+
 ```
-const defaultUrl = ...;
+DEFAULT_URL=...;
 ```
 
 If developing locally and running the Reference Server locally, consider using a tool like `ngrok` to proxy the server endpoints. Once proxied, change `defaultUrl` to the `ngrok` URL endpoints.
```
mhuynh5757, line 53:
```suggestion
If developing locally and running the Reference Server locally, consider using a tool like `ngrok` to proxy the server endpoints. Once proxied, change `DEFAULT_URL` to the `ngrok` URL endpoints.
```

app/README.md
```diff
@@ -70,8 +57,9 @@ See more info about `ngrok` [here](https://ngrok.com/).
 If the Reference Server is deployed to a cloud service (some examples include Google App Engine, Amazon EC2, or similar), then point `defaultUrl` to the deployed service URL.
```
mhuynh5757, line 57:
```suggestion
If the Reference Server is deployed to a cloud service (some examples include Google App Engine, Amazon EC2, or similar), then point `DEFAULT_URL` to the deployed service URL.
```

app/android/app/build.gradle
```diff
@@ -144,7 +147,7 @@ android {
         versionCode 1
         versionName "1.0"
         buildConfigField "boolean", "IS_NEW_ARCHITECTURE_ENABLED", isNewArchitectureEnabled().toString()
-        manifestPlaceholders = [auth0Domain: "{DomainName}", auth0Scheme: "${applicationId}"]
+        manifestPlaceholders = [auth0Domain: project.env.get("DOMAIN_NAME"), auth0Scheme: "${applicationId}"]
```
mhuynh5757, line 150: Awesome! Love this change, won't have to keep a `git stash` around anymore for this haha.

server/README.md
```diff
@@ -16,34 +16,17 @@ See [example.env](example.env) for information on environment variables required
...
-```
-.env
-```
+See [example.env](example.env) for information on Auth0 environment variables required to run the server.
 
 ### Running the Project Locally
 
-* Install dependencies
+- Install dependencies
+
```
mhuynh5757, line 24:
```suggestion
```

mhuynh5757 (changes requested):

PR #57, opened 2023-05-15
[VBLOCKS-1543] : Setup Circle CI with ngrok and e2e login

kpchoy:
[VBLOCKS-1543](https://issues.corp.twilio.com/browse/VBLOCKS-1543)
- Setup ngrok with circleci
- run server in circleci for detox tests

[VBLOCKS-1629](https://issues.corp.twilio.com/browse/VBLOCKS-1629)
- mock login for detox
- navigate to Dialer screen

**Contributing to Twilio**


.circleci/config.yml
```diff
@@ -130,6 +136,53 @@ commands:
...
+          name: "ngrok: install"
+          command: |
+            wget https://bin.equinox.io/c/bNyj1mQVY4c/ngrok-v3-stable-linux-amd64.tgz
+            sudo tar xvzf ngrok-v3-stable-linux-amd64.tgz
+            sudo chmod +x ngrok
+            ./ngrok authtoken ${NGROK_AUTH_TOKEN}
+      - run:
+          name: "ngrok: run"
+          background: true
+          command: ./ngrok http ${PORT}
```
merinsTDL, line 173: The PORT variable is declared in the pipeline environment and defaults to the same RN service used in Android and iOS parallel jobs?

.circleci/config.yml
```diff
@@ -216,6 +274,10 @@ jobs:
       - app-install-dependencies
       - app-install-applesimutils
       - app-install-pods
+      - ngrok-install
```
merinsTDL, line 277: Here it can be a MacOS executor specific step i.e. ngrok-install-macos

merinsTDL (commented):
The default executor(cimg/node:16.19) seems a bit outdated. 
There are tags (cimg/node:lts) or directly current node 18 lts (cimg/node:18.16).

merinsTDL (commented):
The default executor(cimg/node:16.19) seems a bit outdated. 
There are tags (cimg/node:lts) or directly current node 18 lts (cimg/node:18.16).

.circleci/config.yml
```diff
@@ -130,6 +136,53 @@ commands:
...
+            yarn start
+            cd ..
+
+  ##
+  # Ngrok Commands
+  ##
+  ngrok-install:
+    steps:
+      - run:
+          name: "ngrok: install"
```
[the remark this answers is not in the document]
kpchoy: I decided to use homebrew for mac

.circleci/config.yml
kpchoy: We decided it will be easier to do consecutive jobs due to ngrok only allowing 1 tunnel on free accounts

merinsTDL: Right. In general it is better to create an ngrok Twilio token for employees there is option to have access to this feature. Then CCI can pick the ngrok token from environment and will be more stable in longer runs. 
This would ensure 3 things:
1. Specify the domain name for endpoint (could ensure that later it can be known for the app). 
2. Removes1 tunnel limitation. But that applies only to the same host. 
3. Tunnel timeouts and max duration until it rotates the endpoint name is also removed for premium. We had problems with random failures to create ngrok tunnel and job timeouts for free tunnels.

app/src/store/user.mock.ts
```diff
@@ -0,0 +1,153 @@
...
+  { accessToken: string; email: string },
+  void,
+  {
+    state: State;
+    dispatch: Dispatch;
+    rejectValue: { reason: 'LOGOUT_ERROR'; error: any };
+  }
+>('user/logout', async () => {
+  return { accessToken: '', email: '' };
+});
```
charliesantos, line 153: What is the difference between user.ts and user.mock.ts? What is its purpose and when is it used?

scripts/UpdateTwimlApp.js
```diff
@@ -0,0 +1,110 @@
...
+const http = require("node:http");
+const process = require("node:process");
+const NGROK_URL = "http://localhost:4040/api/tunnels";
+const TWIML_APP_URL = (accountSid, appSid) =>
+  `https://api.twilio.com/2010-04-01/Accounts/${accountSid}/Applications/${appSid}.json`;
+const ACCOUNT_SID = process.env.ACCOUNT_SID;
+const TWIML_APP_SID = process.env.TWIML_APP_SID;
+const API_KEY_SID = process.env.API_KEY_SID;
+const API_KEY_SECRET = process.env.API_KEY_SECRET;
+
```
charliesantos, line 12: Please rename this file to all lowercase for consistency.

scripts/UpdateTwimlApp.js
```diff
@@ -0,0 +1,110 @@
...
+    response.setEncoding("utf8");
+    const rawDataChunks = [];
+    response.on("data", (chunk) => {
+      rawDataChunks.push(chunk);
+    });
+    response.on("end", () => {
+      resolve(JSON.parse(rawDataChunks.join()));
+    });
+  };
+}
```
charliesantos, line 34: Handling requests like this tend to be brittle as there are a lot of cases that needs to be handled. The happy path works but this is usually not encourage. A lot of libraries out there are already well tested and made for only handling requests. Why did we not use them (axios/request/superagent)? This file would be so much simpler.

server/example.env
```diff
@@ -32,3 +32,10 @@ AUTH0_AUDIENCE=https://xxxxxxxxxx/api
 # The Domain as a secure URL found in your Auth0 Application settings
 # See https://auth0.com/docs/quickstart/webapp/express/01-login 
 AUTH0_ISSUER_BASE_URL=https://xxxxxxxxx.us.auth0.com/
+
+# CircleCI 
+AUTH0_URL=https://xxxxxxxxx.us.auth0.com/oauth/token
+AUTH0_USERNAME=xxxxxxxxxxxxxx
+AUTH0_PASSWORD=xxxxxxxxxxxxxx
+AUTH0_CLIENT_ID=xxxxxxxxxxxxxx
+AUTH0_CLIENT_SECRET=xxxxxxxxxxxxxx
```
charliesantos, line 41: Do we use these anywhere else other than CircleCI?

server/src/routes/auth0.ts
```diff
@@ -0,0 +1,43 @@
+import { Request, Response } from 'express';
+import { ServerCredentials } from '../common/types';
+import { error } from '../utils/log';
+
+export function createAuth0Route(serverCredentials: ServerCredentials) {
```
charliesantos, line 5: If I understand this correctly, user.mock.ts calls this route to get auth0 token, bypassing the client side auth0 authorization path correct?

We should not add this here since it's only for testing. We don't want customers deploying this route.
Can you make this auth0 request from user.mock.ts instead? You probably need to move the environment variables needed, but I think we already have the tooling for that right?

app/src/store/user.mock.ts
charliesantos: Ah, it seems like this mock file calls /auth0 route. I'm assuming there are duplicate code here right? We only change the part where we call auth0 API to get an auth0 token?

app/e2e/login.test.js
```diff
@@ -18,4 +18,16 @@ describe('Login', () => {
...
+    await element(by.id('login_button')).tap();
+    await expect(element(by.text('Ahoy!'))).toBeVisible();
+    await expect(element(by.text('test_email@twilio.com'))).toBeVisible();
+  });
+
+  it('should navigate to the dialer tab', async () => {
+    await element(by.id('login_button')).tap();
+    await element(by.id('dialer_button')).tap();
+    await expect(element(by.id('dialer'))).toBeVisible();
+  });
```
charliesantos, line 32: This is awesome!

charliesantos (changes requested):
This is awesome @kpchoy ! I'm glad you got it to work. @mhuynh5757 thanks for helping out as well.
I'm ok with the ngrok setup so I think we can close VBLOCKS-1543.
I left comments related to VBLOCKS-1629.

.circleci/config.yml
kpchoy: Yes we will look into upgrading this account to a Twilio Ngrok account in the future

app/src/store/user.mock.ts
kpchoy: Yes `user.mock.ts` is a replacement for `user.ts` when we run the app in `mocked` mode

https://wix.github.io/Detox/docs/guide/mocking/

scripts/UpdateTwimlApp.js
kpchoy: @mhuynh5757 was there a reason for this? Or should I change this request to axios?

server/example.env
kpchoy: No we do not need anywhere else other than CircleCi. I wasn't quite sure if we should add it here or not. But if we don't add it here how will we remember we need these in the CircleCi env?

server/src/routes/auth0.ts
kpchoy: Yes this calls the token route by bypassing the client side auth0 auth path. 

Is it safe to do this in the client side though through `fetch`? I only saw node.js as an option in their docs: https://auth0.com/docs/get-started/authentication-and-authorization-flow/call-your-api-using-resource-owner-password-flow#example-post-to-token-url

app/src/store/user.mock.ts
charliesantos: Is it possible to minimize what we mock here? For example, if we need to change user.ts, we also need to update the mock file and make sure they are in sync. Is it possible to move the block of code related to auth0 only to its own file that way we only mock very minimal code?

scripts/UpdateTwimlApp.js
charliesantos: No need to change now if it adds significant amount of effort. But I let you make the call.

server/example.env
charliesantos: It's ok to put them here if they are required for the server. I was just curious :)

server/src/routes/auth0.ts
charliesantos: It's not safe to put this client side _**iff**_ we are doing this outside of testing environment. However, since we're only putting it in the mock file, we can put this in the client. The node.js option they provided in their doc is just an example. As long as we create the same request from the client, this should work.

app/src/store/user.mock.ts
kpchoy: What it does is it replaces the whole file. To mock only minimal code, I could split up the user.ts file into different files and only mock those files. However that could get messy since we will have to split every function into its own file. What do you think?

mhuynh5757: > Is it possible to minimize what we mock...
Plus one from me on this. One idea is to move all of the auth0 stuff into something like `util/auth.ts`

server/example.env
mhuynh5757: Sorry, does that mean they're only used for testing? If so, they should go into some 1password document or the readme since it's only relevant for us (and I suppose anyone who forks and wants to e2e test as well).

scripts/UpdateTwimlApp.js
mhuynh5757: I wrote it this way for personal use, I didn't want a package just for axios and this one script. Since we're already depending on axios, I'm in favor of using axios (it'll cut the script size down in half honestly).

server/example.env
charliesantos: I believe this is only for documentation. They will go to 1password no matter what.

scripts/UpdateTwimlApp.js
charliesantos: Understood. @kpchoy I'll let you make the call to rewrite this. It's probably not worth it if it's too much work.

app/src/store/user.mock.ts
kpchoy: okay refactored by putting the `thunks` into a `util/auth` file. Then just mocked the `auth` file. I agree it looks cleaner this way

scripts/UpdateTwimlApp.js
kpchoy: i'll create a ticket for the backlog since I am short on time for this ticket

server/example.env
kpchoy: Yes only for testing. I added to `example.env` for documentation, and also added to 1password

app/src/store/user.ts
```diff
@@ -1,8 +1,6 @@
-import { createSlice, createAsyncThunk } from '@reduxjs/toolkit';
-import { getEnvVariable } from '../util/env';
-import Auth0, { type SaveCredentialsParams } from 'react-native-auth0';
-import { State, Dispatch, type AsyncStoreSlice } from './app';
-import { settlePromise } from '../util/settlePromise';
+import { createSlice } from '@reduxjs/toolkit';
+import { type AsyncStoreSlice } from './app';
+import * as Auth from '../util/auth';
```
charliesantos, line 3: Should `Auth` be lowercase? what is our convention if you look at other files?

scripts/UpdateTwimlApp.js
charliesantos: I see that you renamed this now, thanks. Can we delete this one? You have two of these now 😄

charliesantos (changes requested):

scripts/UpdateTwimlApp.js
kpchoy: Strange, my local git only has 1 version

charliesantos (approved):

PR #64, opened 2023-05-31
[VBLOCKS-1715] feat: outgoing call test

kpchoy:
## Submission Checklist

 - [x] [CI pipeline](https://app.circleci.com/pipelines/github/twilio/twilio-voice-react-native-app) is green
 - [ ] Included screenshot if necessary

## Description

[VBLOCKS-1715](https://issues.corp.twilio.com/browse/VBLOCKS-1715)
## Breakdown

- outgoing call UI test
- success case (android only)
- disconnect cases 

## Validation

- [Bulleted summary of validation steps including local and CI pipeline]
- [eg. Added new test]
- [eg. Manually tested with both app and server]

## Additional Notes

[Any additional comments, notes, or information relevant to reviewers.]

**Contributing to Twilio**


CHANGELOG.md
```diff
@@ -3,3 +3,4 @@
 ## App
 * Upgrade React Native from `0.70.6` to `0.70.9`.
   * Potentially fixes iOS builds for newer versions of Xcode.
+* [e2e Tests] outgoing call
```
charliesantos, line 6:
```suggestion
* Added e2e tests for outgoing calls
```

app/e2e/outgoingcall.test.js
```diff
@@ -0,0 +1,82 @@
...
+        await element(by.id('dialpad_button_7')).tap();
+        await element(by.id('dialpad_button_0')).tap();
+        await element(by.id('dialpad_button_3')).tap();
+        await element(by.id('dialpad_button_5')).tap();
+        await element(by.id('dialpad_button_4')).tap();
+        await element(by.id('dialpad_button_1')).tap();
+        await element(by.id('call_button')).tap();
+        await waitFor(element(by.id('active_call'))).toBeVisible();
+        await waitFor(element(by.id('call_status'))).toHaveText('ringing');
+        await waitFor(element(by.id('call_status'))).toHaveText('00:05');
```
charliesantos, line 54: Is this always 5s? Does this test fail sometimes?

app/e2e/outgoingcall.test.js
```diff
@@ -0,0 +1,82 @@
...
+
+    describe('disconnect', () => {
+      it('should disconnect if invalid number', async () => {
+        await element(by.id('dialpad_button_1')).tap();
+        await element(by.id('dialpad_button_2')).tap();
+        await element(by.id('dialpad_button_3')).tap();
+        await element(by.id('call_button')).tap();
+        await waitFor(element(by.id('active_call'))).toBeVisible();
+        await waitFor(element(by.id('call_status'))).toHaveText('ringing');
+        await waitFor(element(by.id('call_status'))).toHaveText('disconnected');
```
charliesantos, line 67: Can you also check that the call gets disconnected right away? It should get disconnected in less than 3s, is that correct?

app/e2e/outgoingcall.test.js
```diff
@@ -0,0 +1,82 @@
...
+        await waitFor(element(by.id('active_call'))).not.toBeVisible();
+      });
+
+      it('should disconnect if invalid Client-ID', async () => {
```
charliesantos, line 71: Same here, we should check that the call has been disconnected right away.

charliesantos (changes requested):
Left some comments. We should also add tests that covers the following:
- check if call is disconnected after clicking end button
- check if call is disconnected after callee disconnects the call or times out
Since this is a timebox task, this can be part of next sprint (new ticket) if you run out of time this sprint.

charliesantos (changes requested):
Left some comments. We should also add tests that covers the following:
- check if call is disconnected after clicking end button
- check if call is disconnected after callee disconnects the call or times out
Since this is a timebox task, this can be part of next sprint (new ticket) if you run out of time this sprint.

app/e2e/outgoingcall.test.js
kpchoy: Since I can only validate UI, the time is the only thing in the UI that verifies a call is happening. I picked 5s since its part of the 10s duration

app/e2e/outgoingcall.test.js
kpchoy: I don't think it can be right away, since the UI still needs time to switch Views. That is why I used a waitFor, since without it the test fails. Where does the 3s come from?

app/e2e/outgoingcall.test.js
kpchoy: Disconnect still doesn't happen right away. We still have the user see texts `ringinging` -> `disconnect` -> back to dial screen.

kpchoy:
> * check if call is disconnected after clicking end button

I tried to do test case `disconnect after clicking end button` , but the test would always fail. I believe this a detox issue, and the issue is still open https://github.com/wix/Detox/issues/1185 

```
No views in hierarchy found matching: (with tag value: is "<Test ID>" and view has effective visibility=VISIBLE
```

kpchoy:
> * check if call is disconnected after callee disconnects the call or times out

Can setup another number and use Twilio <Hangup> verb to end

app/e2e/outgoingcall.test.js
charliesantos: By "right away" I mean, it should not feel like it's stuck. 3s is just an estimate. What we want is, we need to fail this test if it exceeds a certain threshold. For example, if it does not disconnect after 20s, there must be something wrong. 20s is too extreme, so I estimated around 3s. How long do you think we should wait before we consider this as a fail?

app/e2e/outgoingcall.test.js
```diff
@@ -0,0 +1,82 @@
+import { by, element, expect, waitFor, device } from 'detox';
+
+describe('Outgoing Call', () => {
```
charliesantos, line 3: Let's continue the discussion here regarding the following test case, so we can separate the thread

>> - check if call is disconnected after callee disconnects the call or times out

> Can setup another number and use Twilio verb to end

I don't think we need to. We can just wait until the current call times out? It's only 10s long right?

app/e2e/outgoingcall.test.js
kpchoy: I get what you mean now. Okay will add another testcase. 3s sounds like a good time to me 👍

app/e2e/outgoingcall.test.js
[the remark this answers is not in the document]
kpchoy: I had tried various methods earlier in the week, so I didn't think this was possible. But I tried a few more methods yesterday and today, now I have one that works. I have now included a test for your test case.

app/e2e/outgoingCall.test.js
```diff
@@ -0,0 +1,149 @@
...
+          await element(by.id('dialpad_button_1')).tap();
+          await element(by.id('dialpad_button_5')).tap();
+          await element(by.id('dialpad_button_6')).tap();
+          await element(by.id('dialpad_button_7')).tap();
+          await element(by.id('dialpad_button_0')).tap();
+          await element(by.id('dialpad_button_3')).tap();
+          await element(by.id('dialpad_button_5')).tap();
+          await element(by.id('dialpad_button_4')).tap();
+          await element(by.id('dialpad_button_1')).tap();
+          await device.disableSynchronization();
```
charliesantos, line 89: What does this do?

app/e2e/outgoingCall.test.js
```diff
@@ -0,0 +1,149 @@
...
+          await element(by.id('call_button')).tap();
+          await waitFor(element(by.id('end_call_button')))
+            .toExist()
+            .withTimeout(3000);
+          await element(by.id('end_call_button')).tap();
+          await device.enableSynchronization();
```
charliesantos, line 95: What does this do?

charliesantos (approved):
Nice! I'm glad we're able to cover all the cases we talked about. +1 for me.
I also left a couple of questions. You can merge the PR after answering.

app/e2e/outgoingCall.test.js
kpchoy: Detox tries to optimize by "synchronizing" https://wix.github.io/Detox/docs/troubleshooting/synchronization/ . But I think there is some kind of bug triggered by their optimization when trying to click the "end call" button. When in the normal "synchronizing" mode detox cannot find the "end call" button, even though I see it as detox steps through the test case. I found a stackoverflow where someone recommended disabling this "testing opimization" https://stackoverflow.com/questions/59412749/detox-detect-that-element-was-displayed

app/e2e/outgoingCall.test.js
kpchoy: same as above ^

PR #83, opened 2023-07-12
[VBLOCKS-1884] [Server] feat: verify twilio email

kpchoy:
## Submission Checklist

 - [ ] [CI pipeline](https://app.circleci.com/pipelines/github/twilio/twilio-voice-react-native-app) is green
 - [ ] Included screenshot if necessary

## Description

[VBLOCKS-1884](https://issues.corp.twilio.com/browse/VBLOCKS-1884)

## Breakdown

- string matching to verify email
- ENV_VARS as flag to turn on and off the page (same as app flag ENABLE_ABOUT_PAGE)

## Additional Notes

**Contributing to Twilio**


server/src/utils/auth.ts
```diff
@@ -59,3 +60,6 @@ export async function getUserInfo(
     userInfo: response.value.data,
   };
 }
+
+export const verifyTwilioEmail = (email: string): boolean =>
+  email.indexOf('@twilio.com') > -1;
```
charliesantos, line 65: If you change the env to EMAIL_VERIFICATION_REGEX, this can be simplified
```ts
export const verifyTwilioEmail = (email: string, regex: string): boolean =>
  regex && new RegExp(regex).test(email);
```

charliesantos: This way, we also don't have to hardcode `@twilio.com`

charliesantos (changes requested):

server/package.json
```diff
@@ -19,6 +19,7 @@
     "@types/supertest": "^2.0.12",
     "@typescript-eslint/eslint-plugin": "^5.60.1",
     "@typescript-eslint/parser": "^5.60.1",
+    "axios-mock-adapter": "^1.21.5",
```
mhuynh5757, line 22: Opinion of mine, but I don't think we need this package. I think we can just mock `axios` directly, especially since our use-case is very simple. What do you think @charliesantos and @kpchoy?

charliesantos: It's for dev only, makes the test simpler, and not have to do extra work to create the mock, so I'm ok keeping it. We should remove it if we see issues. What are your concerns @mhuynh5757 ?

charliesantos: It's for dev only, makes the test simpler, and not have to do extra work to create the mock, so I'm ok keeping it. We should remove it if we see issues. What are your concerns @mhuynh5757 ?

mhuynh5757: I strongly dislike dependencies, and we're already mocking `axios` in `server/__mocks__`. I don't think it's more than 15 lines to mock `axios.get` ourselves, we just need to modify the `jest.fn` and give it a somewhat proper return value.

charliesantos: Since we're already mocking axios, it's a good reason to not use a separate dependency for mocking. I'm ok removing it.
However, when bringing in dependencies in the future, we should evaluate the dependency objectively, weighing in pros and cons and let the data decide. Just saying we dislike dependencies in general does not help on the decision making process.

mhuynh5757: Ok. Objectively speaking, the `axios-mock-adapter` dependency has few dependencies of its own, just two. And you're correct in that it's only a dev dependency. Furthermore, it seems quite popular, 1m+ DLs a week.

However, I see little benefit of using it since our use-case is so simple, and we only use `axios` in one place. If something breaks with the dependency, even though it's only a dev dependency means that our pipelines will break, our builds won't be green until we address it, etc. It's just more work for us.

charliesantos: +1. I'm convinced.

server/example.env
```diff
@@ -33,4 +33,5 @@ AUTH0_AUDIENCE=https://xxxxxxxxxx/api
 # See https://auth0.com/docs/quickstart/webapp/express/01-login 
 AUTH0_ISSUER_BASE_URL=https://xxxxxxxxx.us.auth0.com/
 
-
+# Note: For internal use only
+ENABLE_ABOUT_PAGE=false
```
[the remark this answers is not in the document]
kpchoy: I thought it would be easier to use the same flag, then we wouldn't need to set two flags for the same feature. But I see the benefit in this, so I will change it.

server/package.json
kpchoy: done

charliesantos (approved):

server/src/__tests__/routes/token.test.ts
```diff
@@ -129,18 +128,19 @@ describe('createTokenRoute()', () => {
     });
 
     describe('Verify Twilio Email', () => {
-      it('returns 401 when About Page is enabled, and email is NOT @twilio.com', async () => {
+      it('returns 401 when EMAIL_VERIFICATION_REGEX is provided, and email is NOT @twilio.com', async () => {
         tokenRoute = createTokenRoute({
           ...mockServerConfig,
-          ENABLE_ABOUT_PAGE: 'true',
+          // eslint-disable-next-line no-useless-escape
```
mhuynh5757, line 134: What does this rule do for this line?

server/src/routes/token.ts
```diff
@@ -37,8 +37,11 @@ export function createTokenRoute(serverCredentials: ServerCredentials) {
...
     if (
-      serverCredentials.ENABLE_ABOUT_PAGE === 'true' &&
-      !verifyTwilioEmail(userInfo.email)
+      typeof serverCredentials.EMAIL_VERIFICATION_REGEX !== 'undefined' &&
+      !verifyTwilioEmail(
+        userInfo.email,
+        new RegExp(serverCredentials.EMAIL_VERIFICATION_REGEX),
+      )
     ) {
       const msg = 'Must be a valid @twilio.com email';
```
mhuynh5757, line 46: Let's reference the regex here, since the regex may not actually be `@twilio.com`.

server/src/utils/auth.ts
```diff
@@ -61,5 +61,5 @@ export async function getUserInfo(
   };
 }
 
-export const verifyTwilioEmail = (email: string): boolean =>
-  email.indexOf('@twilio.com') > -1;
+export const verifyTwilioEmail = (email: string, regex: RegExp): boolean =>
```
mhuynh5757, line 64: Should we name this `twilio` since the regex is adjustable?

server/src/utils/env.ts
```diff
@@ -54,5 +53,11 @@ export function getServerCredentials(): ServerCredentials | undefined {
     return;
   }
 
+  //Optional Credentials
```
mhuynh5757, line 56:
```suggestion
  // Options
```
nit

mhuynh5757 (changes requested):
See above comments.

server/src/routes/token.ts
kpchoy: What should the message be then? 
`Does not match regex` ?

server/src/utils/auth.ts
kpchoy: Just `twilio`? How about `verifyRegex` since the regex is adjustable?

server/src/__tests__/routes/token.test.ts
kpchoy: Actually dont need this anymore, removed.

server/src/utils/auth.ts
mhuynh5757: Yeah sorry haha. I meant to say should this name include twilio

server/src/routes/token.ts
kpchoy: Actually I made it dynamic instead, and then we can show this error message to the user for any email.

mhuynh5757 (approved):

PR #87, opened 2023-07-19
[VBLOCKS-1911] (App) Feat: verify email

kpchoy:
## Submission Checklist

 - [ ] The `CHANGELOG.md` reflects any **feature**, **bug fixes**, or **known issues** made in the source code
 - [ ] [CI pipeline](https://app.circleci.com/pipelines/github/twilio/twilio-voice-react-native-app) is green
 - [x] Included screenshot if necessary

## Description

[VBLOCKS-1911](https://issues.corp.twilio.com/browse/VBLOCKS-1911)

### Breakdown

- Logout User if their email does not match Regex on Server for `getAccessToken` 
- Show error message

## Validation

- Unit test

## Additional Notes

Android:

Screen Shot 2023-07-19 at 3 01 19 PM

iOS:

Screen Shot 2023-07-19 at 4 03 42 PM

## Contributing to Twilio


app/src/screens/SignIn/__tests__/SignIn.test.tsx
```diff
@@ -27,4 +28,22 @@ describe('<SignIn />', () => {
...
+    const initialState = {
+      voice: {
+        accessToken: {
+          status: 'rejected',
+          error: { message: errorMessage },
+        },
+      },
+    };
+    const mockStore = configureStore();
+    wrapper = ({ children }) => (
```
mhuynh5757, line 43: Do we want to put this in a `beforeEach` hook like our other component tests?

app/src/screens/SignIn/index.tsx
```diff
@@ -1,12 +1,13 @@
-import React from 'react';
+import React, { useEffect } from 'react';
```
mhuynh5757, line 1: Nitpick, I prefer us to do `React.useEffect` instead of importing `useEffect` itself.

app/src/screens/SignIn/index.tsx
```diff
@@ -61,10 +64,36 @@ const styles = StyleSheet.create({
...
+    (state: State) => state.voice.accessToken,
+  );
+  const [errorMessage, setErrorMessage] = React.useState<string>('');
+
+  useEffect(() => {
+    if (accessTokenState.status === 'rejected') {
+      setErrorMessage(accessTokenState.error.message);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [accessTokenState.status]);
```
mhuynh5757, line 96: We can move all this logic into the selector callback itself on line `87`. That way we don't have more state than we need and we don't need a potential side-effect from this `useEffect` hook. Also, if we need to use a linting rule to prevent dependency checks, we're probably missing something important.

app/src/screens/StackNavigator.tsx
```diff
@@ -13,13 +13,15 @@ import Busy from '../components/Busy';
...
+    (state: State) => state.voice.accessToken,
+  );
   const user = useSelector((state: State) => state.user);
-
   if (user === null) {
     return <Text>Application not bootstrapped.</Text>;
   }
 
-  if (user?.status === 'pending') {
+  if (user?.status === 'pending' || twilioAccessToken?.status === 'pending') {
```
mhuynh5757, line 24: 👍

app/src/store/__tests__/token.test.ts
```diff
@@ -61,6 +61,7 @@ describe('token store', () => {
     expect(store.getState().voice.accessToken).toEqual({
       status: 'rejected',
       reason: 'USER_NOT_FULFILLED',
+      error: miniSerializeError(null),
```
mhuynh5757, line 64: Can we use something other than `null`?

app/src/store/__tests__/token.test.ts
```diff
@@ -87,12 +88,35 @@ describe('token store', () => {
...
+    });
+  });
+
+  it('rejects if "TOKEN_RESPONSE_NOT_OK" and status is fulfilled', async () => {
+    jest.spyOn(auth0, 'authorize').mockResolvedValueOnce({
+      accessToken: 'test token',
+      idToken: 'test id token',
+    });
+    const errorMessage = 'foobar';
```
mhuynh5757, line 107: Let's use an actual error object. We can pass `"foobar"` as the message for the error but it's closer to the real thing if it's in an error object.

app/src/store/voice/accessToken.ts
```diff
@@ -34,7 +36,10 @@ export const getAccessToken = createTypedAsyncThunk<
 >('voice/getAccessToken', async (_, { getState, rejectWithValue }) => {
   const user = getState().user;
   if (user?.status !== 'fulfilled') {
-    return rejectWithValue({ reason: 'USER_NOT_FULFILLED' });
+    return rejectWithValue({
+      reason: 'USER_NOT_FULFILLED',
+      error: miniSerializeError(null),
```
mhuynh5757, line 41: Do we need this?

app/src/store/voice/accessToken.ts
```diff
@@ -54,14 +59,25 @@ export const getAccessToken = createTypedAsyncThunk<
   }
 
   const tokenResponse = fetchResult.value;
+
+  const tokenTextResult = await settlePromise(tokenResponse.text());
+
+  if (!tokenResponse.ok && tokenTextResult.status === 'fulfilled') {
```
mhuynh5757, line 65: I'm not quite seeing this branch as being necessary. This kind of error should be caught by the branch on line `73`, no?

app/src/store/voice/accessToken.ts
```diff
@@ -91,18 +107,27 @@ export const accessTokenSlice = createSlice({
...
+    builder.addCase(getAccessToken.rejected, (state, action) => {
       switch (action.payload?.reason) {
         case 'USER_NOT_FULFILLED':
           return {
             status: 'rejected',
             reason: action.payload.reason,
+            error: action.payload.error,
           };
         case 'TOKEN_RESPONSE_NOT_OK':
+          state = {
```
mhuynh5757, line 119: The way builder case functions work is that we either mutate the state or return a new state entirely. It seems like we're doing both here.

mhuynh5757 (changes requested):

app/src/store/voice/accessToken.ts
mhuynh5757: Ah, I see now. We want the value of the `text`.

app/src/screens/SignIn/index.tsx
charliesantos: I don't have a preference but I'm curious, do you see issues importing useEffect itself?

charliesantos (commented):
Looks great in general @kpchoy . +1 for me after addressing @mhuynh5757 's comments.

app/src/screens/SignIn/__tests__/SignIn.test.tsx
kpchoy: This file does have the `beforeEach` `store` hook. However, to mock the state, I need to use the `mock-store`.  We can't put it in the `beforeEach` because the `mock state` it will be different for each test case.

app/src/screens/SignIn/index.tsx
kpchoy: Agreed, what is the preference for using `React.` vs `useEffect` itself?

app/src/store/voice/accessToken.ts
kpchoy: Yes I get error:
```
Property 'value' does not exist on type '{ status: "rejected"; reason: unknown; } | { status: "fulfilled"; value: string; }'.
  Property 'value' does not exist on type '{ status: "rejected"; reason: unknown; }'.
```

app/src/store/voice/accessToken.ts
kpchoy: oops old habit

app/src/screens/SignIn/index.tsx
kpchoy: I have an issue when I move this to `useSelector` I get a "too many re-renders" error

app/src/store/voice/accessToken.ts
kpchoy: Without it I have type issues. Might need your help to trouble shoot.

app/src/screens/SignIn/index.tsx
mhuynh5757: No, but that's how it's done in other files. It's more for consistency's sake.

app/src/screens/SignIn/__tests__/SignIn.test.tsx
mhuynh5757: Gotcha. There's actually some substantial work with component testing and using mock stores vs our actual store that I think we need to do. I don't think we can get away with mocking our stores using the mock store library anymore. While working on the audio device component tests, I can't dispatch thunks because the mock store expects plain objects only. Out of scope for this ticket and PR, but something to keep in mind.

mhuynh5757: For now, until we can revisit all of the component tests, I can live with the current implementation in this file.

app/src/screens/SignIn/index.tsx
```diff
@@ -61,10 +64,33 @@ const styles = StyleSheet.create({
...
+    color: '#D61F1F',
+  },
 });
 
 const SignIn: React.FC = () => {
   const dispatch = useDispatch<Dispatch>();
+  const errorMessage = useSelector((state: State) => {
+    if (state.voice.accessToken.status === 'rejected') {
+      switch (state.voice.accessToken.reason) {
+        case 'TOKEN_RESPONSE_NOT_OK':
```
mhuynh5757, line 89: Is this the only failure-state we need to handle?

mhuynh5757: There should probably be a `default` fall-through here or we should be more exhaustive with cases for each reason. Whatever is simpler and still satisfies the compiler. Perhaps some boolean value out of the state on whether or not we should show the error at all too.

app/src/store/voice/accessToken.ts
```diff
@@ -75,7 +89,7 @@ export const getAccessToken = createTypedAsyncThunk<
 
 export type AccessTokenState = AsyncStoreSlice<
   { value: string },
-  GetAccessTokenRejectValue | { error: any }
+  GetAccessTokenRejectValue | { error: any; reason: 'unknown' }
```
mhuynh5757, line 92:
```suggestion
  GetAccessTokenRejectValue | { error: any; reason: 'UNKNOWN_ERROR_OCCURRED' }
```

mhuynh5757: Maybe not this exact wording, but prefer all caps and underscores between words to keep consistency.

mhuynh5757 (changes requested):
Good progress. Take another look at how we're using `miniSerializeError` in this PR though. While it can accept `any`, the underlying mechanism expects an error object or something roughly similar. It will parse `error.message`, `error.stack`, `error.name` etc. out it, so we shouldn't be passing strings or null to it.

app/src/screens/SignIn/__tests__/SignIn.test.tsx
kpchoy: got it. I replaced the `mock-store` with the actual store.

app/src/screens/SignIn/index.tsx
kpchoy: yup only fail-state that we need to handle

mhuynh5757: What happens for `FETCH_ERROR`, `FETCH_TEXT_ERROR`, or `USER_NOT_FULFILLED`? Do we want these errors to effectively be transparent for the end user?

I also think that we should still collapse the branches on line 63 and 71 into one. What about something like
```
if (!tokenResponse.ok) {
  const error = tokenTextResult.status === 'fulfilled' ? new Error(tokenTextResult.value) : tokenTextResult.reason;
  return rejectWithValue({
    ...,
    error: miniSerializeError(error),
  });
}
```
That way we don't have multiple branches repeating some similar stuff. This should work too because if `.text()` fulfills, then it will return a string, which we can put into an error object for serialization. And if it rejects, then the reason is already an error object that can be serialized.

app/src/store/voice/accessToken.ts
```diff
@@ -89,7 +89,7 @@ export const getAccessToken = createTypedAsyncThunk<
 
 export type AccessTokenState = AsyncStoreSlice<
   { value: string },
-  GetAccessTokenRejectValue | { error: any; reason: 'unknown' }
+  GetAccessTokenRejectValue | { error: any; reason: 'UNKNOWN_ERROR' }
```
mhuynh5757, line 92: Nitpick, let's use `UNEXPECTED_ERROR`.

mhuynh5757 (changes requested):

app/src/screens/SignIn/index.tsx
mhuynh5757: Oops, left the comment in the wrong file. It's lines 63 and 71 in the `store/voice/accessToken` file.

kpchoy: done

app/src/store/voice/accessToken.ts
```diff
@@ -60,19 +60,15 @@ export const getAccessToken = createTypedAsyncThunk<
...
-      reason: 'TOKEN_RESPONSE_NOT_OK',
-      statusCode: tokenResponse.status,
-      error: miniSerializeError(tokenTextResult.value),
-    });
-  }
-
   if (!tokenResponse.ok) {
+    const error =
+      tokenTextResult.status === 'fulfilled'
+        ? tokenTextResult.value
```
mhuynh5757, line 66: What does the compiler say the value of `tokenTextResult.value` is here? I think it's a `string`, so we should probably wrap it in an error too.

mhuynh5757 (changes requested):

app/src/store/voice/accessToken.ts
mhuynh5757: Hmm, we shouldn't be returning anything if we're mutating state. It's one or the other.

mhuynh5757: Either we perform the state assignment on line 112, or the return statement on 118. We cannot do both.

mhuynh5757 (changes requested):

mhuynh5757 (approved):

PR #88, opened 2023-07-24
[VBLOCKS-1830] Audio device selection

mhuynh5757:
## Submission Checklist

 - [x] [CI pipeline](https://app.circleci.com/pipelines/github/twilio/twilio-voice-react-native-app) is green
 - [x] Included screenshot if necessary

## Description

This PR implements the store actions and components for proper audio device selection during a call.

### Breakdown

- Add audio device store actions.
- Adjust store to reflect new actions.
- Add components to the active call screen to perform and consume said store changes and actions.
- Adjust hooks for those components.
- Add unit tests for the store and components.

## Validation

- New component and store tests.
- Manual testing.

## Additional Notes

Figma link: https://www.figma.com/file/IFiFQzWEp1xyM2nuqJCkC2/2022-Q2-React-Voice-SDK?type=design&node-id=629-19201&mode=design&t=xDdnuaT6Fub1wU2Y-0

iPhone - Earpiece selected (no bluetooth device connected to iPhone at all)
Screenshot 2023-07-25 at 5 34 09 PM

iPhone - Speaker selected (with or without bluetooth device connected to iPhone)
Screenshot 2023-07-25 at 5 34 17 PM

iPhone - Bluetooth selected (with bluetooth device connected to iPhone)
Screenshot 2023-07-25 at 5 34 24 PM

Android - Earpiece selected

Android - Speaker selected

Note: unable to select earpiece on iPhone when bluetooth device is connected. Is this an iOS limitation? I can select the speaker when there is a bluetooth device, but not earpiece. Calling `.select()` on the earpiece will have the native layer do _something_ and report an audio devices changed event, but the selected device is still bluetooth in that event and indeed the bluetooth device is still relaying audio.

## Contributing to Twilio


charliesantos:
@mhuynh5757 can you please attach some screenshots for each UI state for both platforms?

charliesantos:
@mhuynh5757 do we have figma link handy?

app/src/store/__tests__/audioDevices.test.ts
```diff
@@ -44,73 +56,352 @@ jest.mock('@twilio/voice-react-native-sdk', () => {
...
+        selectedDevice: mockAudioDevices[1],
+      });
+      matchDispatchedActions(dispatchedActions, [
+        audioDevices.updateAudioDevices.pending,
+        audioDevices.updateAudioDevices.fulfilled,
+      ]);
+    });
+
+    it('handles no selected device', async () => {
+      const mockAudioDevices = [
```
kpchoy, line 108: should we move this up one scope so you don't have to keep defining this for every case?

app/src/store/__tests__/audioDevices.test.ts
```diff
@@ -44,73 +56,352 @@ jest.mock('@twilio/voice-react-native-sdk', () => {
...
+        throw error;
+      });
+
+      const result = await store.dispatch(
+        audioDevices.updateAudioDevices({
+          audioDevices: mockAudioDevices as any,
+        }),
+      );
+      if (!audioDevices.updateAudioDevices.rejected.match(result)) {
+        throw new Error('dispatch result should be rejected');
```
kpchoy, line 189: why `throw new Error` instead of use an `expect`?

app/src/store/middleware/log.ts
```diff
@@ -0,0 +1,54 @@
...
+      message += ' ' + JSON.stringify(filteredError, null, 2);
+    }
+
+    console.log(message);
+
+    return next(action);
+  };
+
+/**
+ * Recursively filter out a key from an object.
```
kpchoy, line 37: would an iterative way be better?

charliesantos:
@mhuynh5757 , can we get our Android and iOS members to verify the behavior? Android feels correct to me, but iOS seems odd. Looking at the iOS phone app, it should allow us to select earpiece properly even with bluetooth device connected.

app/__mocks__/@twilio/voice-react-native-sdk.ts
```diff
@@ -72,3 +72,11 @@ export const Call = {
     QualityWarningsChanged: 'qualityWarningsChanged',
   },
 };
+
+export const AudioDevice = {
+  Type: {
+    Earpiece: 'earpiece',
+    Speaker: 'speaker',
+    Bluetooth: 'bluetooth',
```
charliesantos, line 80: Is this always true? Will we ever have more than this?

app/src/screens/ActiveCall/SpeakerButton.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { Image, StyleSheet } from 'react-native';
+import Button, { type Props as ButtonProps } from '../../components/Button';
+
+const SpeakerSource = require('../../../assets/icons/speaker.png');
+const ActiveSpeakerSource = require('../../../assets/icons/speaker-active.png');
+
+export type Props = {
```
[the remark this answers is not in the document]
charliesantos: Do we import this from other files? I cannot find where this is being imported, unless I'm missing it. Also, I realized we follow this same pattern on most files..

app/src/screens/ActiveCall/hooks.ts
```diff
@@ -108,76 +106,86 @@ const useHangup = (
...
-        },
-        false,
-        false,
-      ],
-    )
-    .otherwise(() => [() => {}, true, false]);
+/**
+ * Return type of the audio device hook.
+ */
+type AudioDevicesHookReturnType = {
```
charliesantos, line 112: Too verbose and the name does not really describe the members of this type. How about `AudioDevicesHook` or `AudioDevicesHandlers`?

app/src/store/middleware/log.ts
```diff
@@ -0,0 +1,54 @@
...
+      const error = action.payload || action.error;
+
+      const filteredError = logStackTrace
+        ? error
+        : filterOutKey('stack', error);
+
+      message += ' ' + JSON.stringify(filteredError, null, 2);
+    }
+
+    console.log(message);
```
charliesantos, line 31: What does this look like from the console? Would you be able to see where the log is coming from, or would this only show that all logs are coming from `log.ts`?

app/src/screens/ActiveCall/SpeakerButton.tsx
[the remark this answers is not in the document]
mhuynh5757: We don't import these in other files yet. In the past with other components I've found it necessary sometimes to be able to get the Props of another component (especially when extending the component) so I do it by default. If we want to remove the exports we can, but they don't hurt.

I also find that exporting as `Props` makes more since than `SpeakerButtonProps` because you import like so
```
import { Props } from './SpeakerButton';
```
It's very clear it's props for the speaker button, and having it named `SpeakerButtonProps` is redundant. If there's a naming clash (like you need to import props from multiple components) you can just do this
```
import { Props as SpeakerButtonProps } from './SpeakerButton';
```

app/src/screens/ActiveCall/hooks.ts
```diff
@@ -108,76 +106,86 @@ const useHangup = (
...
+    return selectedType === TwilioAudioDevice.Type.Bluetooth
+      ? selectEarpiece
+      : selectBluetooth;
+  }, [selectEarpiece, selectBluetooth, selectedType]);
+
+  if (
+    typeof activeCall === 'undefined' ||
+    audioDevices.status !== 'fulfilled'
+  ) {
+    return {};
```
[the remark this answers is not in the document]
mhuynh5757: No, we shouldn't throw errors in hooks, instead the hook-managed state should reflect that. In this case, the handlers will be undefined since we're returning an empty object. And the component will be disabled since the handlers are undefined.

app/src/store/__tests__/audioDevices.test.ts
mhuynh5757: We could, but we also modify the devices slightly with every test-case.

app/src/store/__tests__/audioDevices.test.ts
mhuynh5757: Using an expect would look something like this
```
expect(audioDevices.updateAudioDevices.rejected.match(result)).toBe(true);
```
If this expectation fails, then the failure message in jest would just be something simple like
```
expected true, got false
```
or something, which isn't the most helpful error message in the world. If we throw an error with a specific error message, then I think it helps debugging unit tests.

app/src/store/middleware/log.ts
mhuynh5757: To be honest, probably not but I also can't think of a way to iteratively remove all keys from nested objects.

app/__mocks__/@twilio/voice-react-native-sdk.ts
mhuynh5757: This is it according to the native layers.

app/src/screens/ActiveCall/SpeakerButton.tsx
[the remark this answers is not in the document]
kpchoy: I know they don't hurt, but usually when I see exports I expect them to be used somewhere else.

kpchoy (approved):

app/src/screens/ActiveCall/SpeakerButton.tsx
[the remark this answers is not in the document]
charliesantos: I think we should not export things that are not used outside of the file. We can export it once we need it. We also missed this last time. It seems we are already exporting a bunch of these things that don't need to be exported. I'm ok to keep it as is but we need to have a todo or ticket to stop exporting these.

app/src/store/middleware/log.ts
charliesantos: Weird, no line numbers or source file info?

mhuynh5757: Not in the terminal, no.

app/src/screens/ActiveCall/SpeakerButton.tsx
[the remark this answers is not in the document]
mhuynh5757: Ok, but curious as to why it's an issue. We don't export it out of the project for the consuming application, so it's not visible to the end developer. And it's a type, so it's not mutable or anything dangerous.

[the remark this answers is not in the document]
mhuynh5757: But I do know that it's best practice typically to not needlessly export things, so I will not continue this pattern going forward.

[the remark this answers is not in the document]
charliesantos: One, just good practice. If you don't need to export, don't export, preventing future mistakes in case we need to encapsulate something. The other thing is, just like what kevin mentioned, it adds confusion. If you see an export keyword here, I would assume it's being used somewhere so changing it might be more risky since I wouldn't want other files to break.

charliesantos (approved):

charliesantos (approved):

